PR #3, opened 2017-04-02
API to wake-up from deepsleep using gpio

jagruthp:
Added API to wake-up from deepsleep using gpio pins.


danicampora:
@jagruthp thanks for your contribution. Looks good but I have a few remarks that should be easy for you to work on:

- Use keyword arguments, see an example here: https://github.com/pycom/pycom-micropython-sigfox/blob/master/esp32/mods/machspi.c#L350
- Instead of simply printing a message to notify about an error, you should raise an exception with a message, example: https://github.com/pycom/pycom-micropython-sigfox/blob/master/esp32/mods/machpin.c#L331
- Finally, I'd remove the timeout parameter, and won't start the deep sleep process here. Inside the machine module we have a function that does already. The pin module should simply enable the wake_up source.

jagruthp:
@danicampora  - I have made the changes according to your remarks. Also, I misunderstood what was required from the mail. I thought I had to create an API to set the trigger of the pin and also put the system to deepsleep. But now since only the triggers are set, I have changed the name of the API to wakeup_trigger as this would make more sense than the name deepsleep.

jagruthp:
Yes this works. I tested it on the board before I committed it. I saw the examples from the SPI class. I also saw the methods used in the Pin class. 

I see what you mean by using MP_DEFINE_CONST_FUN_OBJ_KW but the API works because my args are of type MP_ARG_INTand not MP_ARG_OBJ/MP_ARG_KW.  In mp_arg_parse_all() in argcheck.c, after executing till line 84, line 106 gets executed next which gives me the int value of the entered trigger level.

Could you please test it out at your end and let me know if I am wrong ? If you still want me to use MP_DEFINE_CONST_FUN_OBJ_KW, I can do that and commit it by EOD.

danicampora:
@jagruthp OK, yes that works, but for consistency please change it to MP_DEFINE_CONST_FUN_OBJ_KW. Thanks :-)

jagruthp:
@danicampora - I've modified the API to use MP_DEFINE_CONST_FUN_OBJ_KW :-)

PR #12, opened 2017-05-28
Adding support for selecting device class for LoRaWAN

husigeza:
- Bugfix: tx_retries is incorrectly initialized
- New: new parameter when creating a new LoRa class and calling init method: device_class: CLASS_A and CLASS_C is accepted
- New method for LoRa class: get_device_class, returns with the current device class (as a string) or with a message if the mode is not LoRaWAN.


husigeza:
Raising a ValueError if the device class has not been configured before calling get_device_class() method.

PR #13, opened 2017-05-29
mods/modlora.c: Setting the maximum payload for LoRaRAW packets explicitly to be able to receive packets bigger than 64 bytes.

husigeza:
Could not receive packets with size bigger than 64 bytes in LoRaRAW mode. In LoRaWAN mode it works because the Maximum Payload Length of the packets are set before every transmission but in LoRaRAW it was not set.

### DIFF
--- a/esp32/mods/modlora.c
+++ b/esp32/mods/modlora.c
@@ -72,6 +72,8 @@
 #define LORA_TX_TIMEOUT_MAX                         (9000)      // 9 seconds
 #define LORA_RX_TIMEOUT                             (0)         // No timeout
 
+#define LORA_MAX_PACKET_SIZE                        (255)
+
 // [SF6..SF12]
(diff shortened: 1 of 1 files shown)


esp32/mods/modlora.c
```diff
@@ -72,6 +72,8 @@
 #define LORA_TX_TIMEOUT_MAX                         (9000)      // 9 seconds
 #define LORA_RX_TIMEOUT                             (0)         // No timeout
 
+#define LORA_MAX_PACKET_SIZE                        (255)
+
```
danicampora: We already have:

```
#define LORA_PAYLOAD_SIZE_MAX                                   (255)
```

in modlora.h, but the rest is great. Good find :-)

husigeza: Corrected in next commit :)

danicampora:
Merged, thanks!

PR #25, opened 2017-06-23
space in function parameter name "number 2"

livius2:
trivial correct of parameter name 
from "number 2" to "number2"


danicampora:
Merged internally. Thanks @livius2 !

PR #28, opened 2017-06-29
esp32/lore/spi-board.c: Using of macros with prefix DPORT_

husigeza:
esp32/mods/machspi.c: Using of macros with prefix DPORT_


danicampora:
Merged, thanks!

PR #31, opened 2017-07-01
Add

hcfman:
Added support for optional argument "dr" in the join command for Lora

### DIFF
--- a/esp32/mods/modlora.c
+++ b/esp32/mods/modlora.c
@@ -205,6 +205,7 @@ typedef struct {
           uint8_t           DevEui[8];
           uint8_t           AppEui[8];
           uint8_t           AppKey[16];
+          uint8_t           data_rate;
       } otaa;
 
       struct {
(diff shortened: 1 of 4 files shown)


hcfman:
Tested against values of dr from 0-8 (Of which 0-6 are valid and choosing other values will result in the original behavior).

hcfman:
Oops! I thought that close was a submit. Re-opened. I assume that this means the request is submitted now?

danicampora:
@hcfman nice!! Well done :-). I only have a small remark related to the **data_rate_to_counter_map** array.

hcfman:
Let me guess. The size doesn't add up? Normally I would create that array with [] and let the compiler fill it. However this compiler was generating errors saying that I was indexing out of bounds. Odd, because it was indexed with a variable, it shouldn't know about that at compile time. Also if I set an explicit size the exact size of the array it size still complained. I had to make it one size bigger than the actually required. Very strange compiler behavior. But ok, that little but of strangeness got rid of the bizarre compiler error. Looks like it's trying to be a very smart little compiler but not quite smart enough.

- Cheers
Kim

PR #35, opened 2017-07-12
Added ability to set LoRa WAN (packet) Port

ScR4tCh:
Added ability to change LoRa WAN messages port by setting  socket.SO_PORT with socket.setsockopt.
Example:
`s = socket.socket(socket.AF_LORA, socket.SOCK_RAW)`
`s.setsockopt(socket.SOL_LORA, socket.SO_PORT, 222)`


danicampora:
@ScR4tCh thanks for your contribution, however setting the port for LoRaWAN is already possible using the **bind()**: https://docs.pycom.io/chapter/firmwareapi/pycom/network/lora.html

socket.bind(port_number)

https://github.com/pycom/pycom-micropython-sigfox/blob/master/esp32/mods/modlora.c#L1860

ScR4tCh:
@danicampora 
I see. I was a bit confused about bind, as I'd normally (speaking about tcp/udp) use socket.connect(add,port) and only use bind for server-sockets. I should have tried this in advance ;).
Thx for the quick reply.

PR #45, opened 2017-08-01
Reimplementing alarm_heap management to make it simpler

husigeza:
esp32/mods/mactimer_alarm.c: Reimplementing alarm_heap management


danicampora:
Merged long ago. Thanks!

Issue #17, opened 2017-06-07
Build failure on OSX


jeremyherbert:
When building for LoPy: `make BOARD=LOPY LORA_BAND=USE_BAND_915 TARGET=app`

I needed to do this first: https://github.com/micropython/micropython/commit/326343feebdd1ad990864cc830a30b6a44d44918

jeremyherbert:
I'm now stuck at this error:

```
CC mods/machtimer_alarm.c
In file included from /Users/jeremy/Code/pycom-micropython-sigfox/pycom-esp-idf/components/soc/esp32/include/soc/soc.h:20:0,
                 from /Users/jeremy/Code/pycom-micropython-sigfox/pycom-esp-idf/components/driver/include/driver/timer.h:19,
                 from mods/machtimer_alarm.c:6:
/Users/jeremy/Code/pycom-micropython-sigfox/pycom-esp-idf/components/soc/esp32/include/soc/io_mux_reg.h: In function 'PIN_PULLUP_DIS':
/Users/jeremy/Code/pycom-micropython-sigfox/pycom-esp-idf/components/esp32/include/esp_assert.h:23:28: error: first argument to '__builtin_choose_expr' not a constant
             _Static_assert(__builtin_choose_expr(__builtin_constant_p(CONDITION), (CONDITION), 1), #MSG);   \
                            ^
/Users/jeremy/Code/pycom-micropython-sigfox/pycom-esp-idf/components/soc/esp32/include/soc/soc.h:138:38: note: in expansion of macro 'TRY_STATIC_ASSERT'
 #define ASSERT_IF_DPORT_REG(_r, OP)  TRY_STATIC_ASSERT(!IS_DPORT_REG(_r), (Cannot use OP for DPORT registers use DPORT_##OP));
                                      ^
/Users/jeremy/Code/pycom-micropython-sigfox/pycom-esp-idf/components/soc/esp32/include/soc/soc.h:169:13: note: in expansion of macro 'ASSERT_IF_DPORT_REG'
             ASSERT_IF_DPORT_REG((_r), REG_CLR_BIT);                                                                    \
             ^
/Users/jeremy/Code/pycom-micropython-sigfox/pycom-esp-idf/components/soc/esp32/include/soc/io_mux_reg.h:96:5: note: in expansion of macro 'REG_CLR_BIT'
     REG_CLR_BIT(PIN_NAME, FUN_PU);
     ^
[...]

PR #50, opened 2017-08-10
Let make target flash respect TARGET parameter.

eqrx:
As stated in issue #17 the current boot part of the esp32 image does not work. While this is a problem I need a way to just flash the "app" part of the build. This PR changes the makefile to respect the TARGET parameter for the flash target so I do not need the spin up `esptool` explicitly.
`make BOARD=WIPY TARGET=app flash` will only flash the app part, `make BOARD=WIPY TARGET=boot flash` will flash boot and the partition table, `make BOARD=WIPY flash` will flash both. Everything else remains unchanged

### DIFF
--- a/esp32/application.mk
+++ b/esp32/application.mk
@@ -393,6 +393,15 @@ $(BUILD)/esp32_out.ld: $(ESP_IDF_COMP_PATH)/esp32/ld/esp32.ld sdkconfig.h
 	$(Q) $(CC) -I. -C -P -x c -E $< -o $@
 endif
 
+ifeq ($(TARGET), app)
+flash: $(APP_BIN)
+	$(ECHO) "Flashing app"
+	$(Q) $(ESPTOOLPY_WRITE_FLASH) $(APP_OFFSET) $(APP_BIN)
+else ifeq ($(TARGET), boot)
+flash: $(BOOT_BIN)
+	$(ECHO) "Flashing boot and partition"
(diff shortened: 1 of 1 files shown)


eqrx:
Thank you :) I will rebase the PR onto the master branch regularly. Feel free to use it.

nevercast:
@eqrx @Xykon As stated in this issue, the boot part does not work. It was mentioned that #17 says why, but I seem to have missed that detail, isn't #17 about compiling the app image? What is wrong with the boot image and is there already an issue open about it?

Cheers, Josh.

eqrx:
Hello @nevercast 

I had troubles flashing my firmware (I just use the configuration from this repository plus some frozen modules, fyi: https://github.com/eqrx/pycom-micropython-sigfox) and found issue #17 where @jeremyherbert stated that the flashing of the full image produced by this repo will result in a not working esp32. After power connect the device displays some generic information via UART but simply stops before presenting a python promt without any output. Since the issue is four month old I do not have the setup to reproduce the error.

I was hoping that the pycom maintainers would react and take a closer look at the problem so I modified the makefile to flash just the app part for the time being.

nevercast:
@eqrx Over the last couple days I've managed to get my script flashing the partitions, bootloader and application now and have it booting on latest IDF and Firmware + my frozen modules.

I'm not sure what has changed, but before it didn't work and now it does... I'll return if it breaks again or I'm summoned.

PR #59, opened 2017-08-27
mpconfigport.h: Enable running pre-compiled scripts from the file system

robert-hh:
This feature is supported in the code, but not enabled yet. Enabling it
allows to run code pre-compiled with mpy-cross without the need to embed it
into the formware as frozen bytecode. Running bytecode from the file system
saves the compiling phase, allowing larger scripts and giving faster load time.

### DIFF
--- a/esp32/mpconfigport.h
+++ b/esp32/mpconfigport.h
@@ -42,6 +42,7 @@
 #include <stdint.h>
 
 // options to control how Micro Python is built
+#define MICROPY_PERSISTENT_CODE_LOAD        (1)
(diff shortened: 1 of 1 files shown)


danicampora:
Merged, thanks!

robert-hh:
Hello @danicampora : I prepared another PR on ESP32 which adds the method machine.setinterrupt() controlling the interrupt char for keyboard interrupt. If I attempt to do this new PR, it will be submitted together with this one. Is that a problem?

danicampora:
@robert-hh that's OK, go ahead and submit it like that. Thanks!

PR #60, opened 2017-08-27
ESP32: Add method machine.setinterrupt() controlling the interrupt char

robert-hh:
Set the character which interrupts keyboard input of methods like sys.stdin and input.
machine.setinterrupt(c) set the interrupt character for the keyboard functions.
machine.setinterrupt(-1) practically disables keyboard  interrupt,
machine.setinterrupt(3) sets it back to Ctrl-C.


dpgeorge:
Please consider following upstream and using `micropython.kbd_intr()` as per http://docs.micropython.org/en/latest/pyboard/library/micropython.html#micropython.kbd_intr

robert-hh:
@dpgeorge That's a very good hint. It keeps the ports together. I just remembered the function from PyBoard, where it is located in the USV_VCP module.

PR #61, opened 2017-08-28
 ESP32: Enable method micropython.kbd_intr()

robert-hh:
This method allows setting the keyboard interrupt character.
micropython.kbd_intr(-1) disables keyboard interrupts, while
micropython.kbd_intr(3) sets it back to Ctrl-C


robert-hh:
I added code for dac.tone to handle the second parameter of the method, which was documented but not implemented.

robert-hh:
I'll re-issue separate PRs for the kbd_intr() and DAC.tone() fixes.

PR #65, opened 2017-09-01
esp32: Add code for the second parameter of DAC.tone()

robert-hh:
This parameter specifies the level, which is already documented, but was
not implemented. Suitable values:
0: Tone level 3.02 Vpp at 560 Ohm load, about 0 dbV
1: Tone level 1.54 Vpp, about -6 dBV
2: Tone level 0.8 Vpp, about -12 dBV
3: Tone level 0.4 Vpp, about -18 dBV

Adding a small load is reccomended, since that also suppresses
DAC reloading spikes.


oligauc:
@robert-hh Thanks

robert-hh:
@oligauc  What? The comments says: "Closed with unmerged commits"

oligauc:
Hello Robert, Your patch will be available in the next releases. It must undergo regression and unit tests before it makes it to the official repo.

Issue #63, opened 2017-08-31
ustruct Guru Meditation Error crash and blocks the device


syndycat:
LopY 1.7.8.b1

This type of crash  blocks the device. 

Simple to reproduce. Type in console:

a= 0xaabbccdd
a.to_bytes(4)
b'\xdd\xcc\xbb\xaa'

import ustruct
ustruct.pack('4b',a)

Guru Meditation Error of type LoadProhibited occurred on core  0. Exception was unhandled.
Register dump:
PC      : 0x400ee210  PS      : 0x00060030  A0      : 0x800ee403  A1      : 0x3ffc6830  
A2      : 0x3ffc6af0  A3      : 0x3ffc6841  A4      : 0x00000001  A5      : 0x3ffe83f2  
A6      : 0x000000fb  A7      : 0xffffffff  A8      : 0x00000000  A9      : 0x0000ccdd  
A10     : 0x00000002  A11     : 0x3ffe83f0  A12     : 0x0000ccdd  A13     : 0x3ffe83ef  
A14     : 0x00000010  A15     : 0x3ffe83f1  SAR     : 0x0000001e  EXCCAUSE: 0x0000001c  
EXCVADDR: 0x00000004  LBEG    : 0x4000c28c  LEND    : 0x4000c296  LCOUNT  : 0x00000000  

Backtrace: 0x400ee210:0x3ffc6830 0x400ee400:0x3ffc6850 0x400f5ef0:0x3ffc6870 0x400f77ae:0x3ffc68b0 0x400f7801:0x3ffc68f0 0x400f078b:0x3ffc6920 0x400ecc5d:0x3ffc6950 0x400eccc5:0x3ffc6970 0x400f883d:0x3ffc6990 0x400f0710:0x3ffc6a30 0x400ecc5d:0x3ffc6ab0 0x400ecc8a:0x3ffc6ad0 0x400d9953:0x3ffc6af0 0x400d9bf4:0x3ffc6b90 0x400d8c28:0x3ffc6bd0

Guru Meditation Error of type LoadProhibited occurred on core  0. Exception was unhandled.
Register dump:
PC      : 0x4008a199  PS      : 0x00060033  A0      : 0x80088fd6  A1      : 0x3ffc6470  
A2      : 0x3ffc64c0  A3      : 0x3ffc6490  A4      : 0x00000000  A5      : 0x3ffc483c  
[...]

PR #66, opened 2017-09-02
ustruct: Pull in ustruct module from the Micropython.org tree

robert-hh:
This addresses the issue #63, fixed by Damien George @dpgeorge
(PR just in case no one else rang the bell.)

### DIFF
--- a/py/modstruct.c
+++ b/py/modstruct.c
@@ -1,5 +1,5 @@
 /*
- * This file is part of the Micro Python project, http://micropython.org/
+ * This file is part of the MicroPython project, http://micropython.org/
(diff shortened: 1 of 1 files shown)


danicampora:
@robert-hh thanks, will merge!

oligauc:
@robert-hh Thanks

robert-hh:
Do you plan to merge #64 and #65 too? 
#64 would serve the needs of at least one other person who had troubles with non-transparent sys.stdin().
#65 implements a functionality which is already documented. So it has either to be implemented, or the documentation has to be changed.

PR #74, opened 2017-09-28
Return None for unknown nvs_get keys

dbrgn:
Currently the `nvs_get` function will initialize unknown values to 0. This results in two problems:

* Every read on an unknown key will result in a write.
* It is not possible to discern a value that has been set to 0, and a value that does not exist.

By returning `None` for unknown values, discerning missing and 0 values becomes possible.

Note: There are some additional problems with the nvs functionality:

* NVRAM values cannot currently be erased. That's a problem because NVRAM could fill up, rendering the `nvs_*` functions useless without compiling custom firmware
* The error handling is lacking. For example, according to [the docs](https://www.tiny-dev.com/esp32/api/nvs_flash.html#_CPPv210nvs_set_i810nvs_handlePKc6int8_t), there are 5 different error codes for NVRAM write problems. It would be good if these were passed to the user, probably via exceptions.
* Note also that right now (even with this fix) a uninitialized variable will be passed to `mp_obj_new_int` if `nvs_get_u32` returns an error code other than `ESP_ERR_NVS_NOT_FOUND`.

Note: This would of course have to be mentioned in the release notes, since it's a breaking change (the return type is changed).

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,15 +135,15 @@ the Espressif website:
 
 - for 64-bit Linux::
 
-    https://dl.espressif.com/dl/xtensa-esp32-elf-linux64-1.22.0-59.tar.gz
+    https://dl.espressif.com/dl/xtensa-esp32-elf-linux64-1.22.0-61-gab8375a-5.2.0.tar.gz
 
 - for 32-bit Linux::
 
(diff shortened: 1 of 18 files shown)


danicampora:
Merged long ago, thanks. New API functions were added to erase the NRAM area in case it gets full.

PR #80, opened 2017-10-09
Fixes an SSL / TLS socket blocking problem

peterson79:
I don't know why the "MBEDTLS_ERR_SSL_TIMEOUT" is thrown after a connection has succeeded.  However, the "do nothing" loop in combination with this error in "wlan_socket_recv" was causing the socket to block even when the timeout was set to 0.  Catching this error and returning zero bytes on a non-blocking ssl socket seems to work.  I tested this quickly with a self signed cert on a node.js server...

If this seems too "hacky" then mbedtls needs debugged and modified appropriately.  The mbedtls code is wrapped up in a library at the moment so the source is not readily available in the tree...


danicampora:
Looks good, thanks!

danicampora:
Merged!

PR #100, opened 2017-11-20
Added hostname configuration to WLAN module

illysky:
Added hostname when creating WLAN STA or AP object. Useful for getting the device IP address from router attached devices. Overrides “espressif” hostname.

eg,
from network import WLAN
w = WLAN(mode=WLAN.STA, hostname=“Pycom”) 

I’ve limited the hostname to 16 characters, not sure what the limit should be off top of my head. 

Cheers

### DIFF
--- a/esp32/mptask.c
+++ b/esp32/mptask.c
@@ -410,8 +410,7 @@ STATIC void mptask_update_lpwan_mac_address (void) {
 #endif
 
 STATIC void mptask_enable_wifi_ap (void) {
-    wlan_setup (WIFI_MODE_AP, DEFAULT_AP_SSID, WIFI_AUTH_WPA2_PSK, DEFAULT_AP_PASSWORD,
-                DEFAULT_AP_CHANNEL, ANTENNA_TYPE_INTERNAL, true);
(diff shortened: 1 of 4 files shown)


esp32/mptask.c
```diff
@@ -410,8 +410,7 @@ STATIC void mptask_update_lpwan_mac_address (void) {
 #endif
 
 STATIC void mptask_enable_wifi_ap (void) {
-    wlan_setup (WIFI_MODE_AP, DEFAULT_AP_SSID, WIFI_AUTH_WPA2_PSK, DEFAULT_AP_PASSWORD,
-                DEFAULT_AP_CHANNEL, ANTENNA_TYPE_INTERNAL, true);
+    wlan_setup (WIFI_MODE_AP, DEFAULT_AP_SSID, WIFI_AUTH_WPA2_PSK, DEFAULT_AP_PASSWORD, DEFAULT_AP_CHANNEL, ANTENNA_TYPE_INTERNAL, true, "illysky");
```
ep1cman: the `illysky` default should be changed to something more appropriate to all users

livius2:
why hostname have limit to 16 chars?
And if providing function, then better will be **wlan_validate_hostname** not only wlan_validate_hostname_len...
but i do not know if this is not validated already inside tcpip_adapter_set_hostname

esp32/mptask.c
illysky: Ah yes! Sorry forgot about that. Perhaps “Pycom” or “MicroPython”.

PR #109, opened 2017-12-21
sflash_diskio.c: Use spi_flash_get_chip_size() to determine the flash size

robert-hh:
This is the direct method of using the flash size as reported by the spi_flash_get_chip_size() call instead of assuming the flash size by the chip revision number.


danicampora:
@robert-hh nice one, thanks!

danicampora:
Merged with master, thanks.

PR #110, opened 2017-12-21
modpycom.c/pulses_get(): prevent lock-up at long pulses

robert-hh:
pulses_get() gets into a kind of lock state, if pulses with a duration of > 20000µs are applied and the time out was set to an appropriate large value, like 30000. It does not simply return wrong values, but from one "overflow" event on it returns the value pair 19879, 0 irrespective of the signal applied to the input. Only hard reset fixes that. 
This PR  results in a more robust behavior. For pulses of > 32768µs a wrong value will be reported, but after returning to shorter pulses, these will be reported correctly.


danicampora:
Merged with master, thanks!

robert-hh:
B.T.W.: Merry Christmas

danicampora:
Merry Christmas!! :-)

Issue #113, opened 2017-12-25
utime: ticks_diff() raises an exception with large values returned by ticks_us() or ticks_ms()


robert-hh:
The functions utime.ticks_us() and utime.ticks_ms() may return values, which then cannot be used as input for utime.ticks_diff(). Example:
```
>>> time.ticks_us()
3636776263
>>> time.ticks_diff(time.ticks_us(), time.ticks_us()+1)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
OverflowError: overflow converting long int to machine word
>>> 
```
The reason is, that the return values are not stripped to values which can be representef by a MP_SMALL_INT, or at least 31 bit.
Edit: Funny enough, the utime module is mostly implemented twice, in esp32/mods/modutime.c and in py/extmod/mphal_utime.c, where the latter seems to have proper value/overflow handling.

robert-hh:
Made a PR attemting to fix that issue.

robert-hh:
This simple script will trigger an exception latest after about 35 minutes.
```
import utime

new = utime.ticks_us()
while True:
    utime.sleep(1)
    old = new
    new = utime.ticks_us()
    print(new, utime.ticks_diff(old,new))
```

robert-hh:
Actually the minimal change to fix that is to modify ticks_diff() as:
```
STATIC mp_obj_t time_ticks_diff(mp_obj_t start_in, mp_obj_t end_in) {
    int32_t start = mp_obj_get_int_truncated(start_in);
    int32_t end = mp_obj_get_int_truncated(end_in);
    return mp_obj_new_int((end - start));
}
STATIC MP_DEFINE_CONST_FUN_OBJ_2(time_ticks_diff_obj, time_ticks_diff);
```
This change will also return a reasonable behavior, if the arguments of the call are swapped - sign change. The only deviating result is achieved with ticks_diff(0x80000000, 0), and ticks_diff(0, 0x8000000), which return the same value -2147483648

robert-hh:
Thanks for merging.

PR #115, opened 2017-12-25
modutime.c: Ensure proper smallint values on overflow

robert-hh:
This fixes ensure, that the return values of ticks_ms(), ticks_us() and ticks_cpu() fit into a MP_SMALL_INT. That way, they can be fed into ticks_diff() without the risk of an OverflowError (see issue #113). As a side effect, the result of ticks_diff(a, b) is always defined as a reasonable value even if a is a time after b (a > b). In that case, the value is negative. The implementation just copies the mechanism from py/extmod/mphal_utime.c.
Second change: Add the method ticks_add().

### DIFF
--- a/esp32/mods/modutime.c
+++ b/esp32/mods/modutime.c
@@ -155,41 +155,31 @@ STATIC MP_DEFINE_CONST_FUN_OBJ_1(time_sleep_us_obj, time_sleep_us);
 STATIC mp_obj_t time_ticks_ms(void) {
     struct timeval tv;
     gettimeofday(&tv, NULL);
-    return mp_obj_new_int_from_uint(((tv.tv_sec * 1000) + tv.tv_usec / 1000)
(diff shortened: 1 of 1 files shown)


robert-hh:
Letting ticks_ms() and ticks_us() return 64 bit values would not change that behavior, but then ticks_diff() would be obsolete, because if can re replaced by calculating the difference in Python. The code for that could be:
```
STATIC mp_obj_t time_ticks_ms(void) {
    return mp_obj_new_int_from_ll(system_get_rtc_time() / 1000);
}
STATIC MP_DEFINE_CONST_FUN_OBJ_0(time_ticks_ms_obj, time_ticks_ms);

STATIC mp_obj_t time_ticks_us(void) {
    return mp_obj_new_int_from_ll(system_get_rtc_time());
}
STATIC MP_DEFINE_CONST_FUN_OBJ_0(time_ticks_us_obj, time_ticks_us);
```

PR #116, opened 2017-12-29
extmod/utimeq.c: queue module supporting the asyncio lib for cooperative multitasking

robert-hh:
Both the module utimeq.c and the asyncio lib are made by @pfalcon (Paul Sokolovsky). They create support for cooperative multitasking, in contrast to exhaustive multitasking offered by the _thead module. This PR just adds the utimeq.c module to the set of extern modules and aligns the various settings. The asyncio libs themselves requires slight changes too. The major change is caused by the diverging utime module. Some of that is addressed by the PR #115 to the utime module.

### DIFF
--- a/extmod/modutimeq.c
+++ b/extmod/modutimeq.c
@@ -1,0 +1,230 @@
+/*
+ * This file is part of the MicroPython project, http://micropython.org/
+ *
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2014 Damien P. George
+ * Copyright (c) 2016-2017 Paul Sokolovsky
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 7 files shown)


husigeza:
Hello,
What is the purpose of having the utimeq module alone ? Isn't that needed by the cooperative multitasking ?
Geza

danicampora:
I believe utimeq is needed by these libraries: https://github.com/micropython/micropython-lib/tree/master/asyncio_slow

danicampora:
Sorry, it's actually needed here: https://github.com/micropython/micropython-lib/tree/f0ed97ad23b1ee5f28fc214e264a4740fd32d32b/uasyncio.core

husigeza:
Yes that's true, I missed it. The change looks good except it crashes if the size of the list passed to the pop() function is smaller than 3 because the mp_raise_TypeError() function is called with NULL value. It should be changed to something: "The list does not exist or too small!"

extmod/modutimeq.c
```diff
@@ -0,0 +1,230 @@
...
+STATIC MP_DEFINE_CONST_FUN_OBJ_VAR_BETWEEN(mod_utimeq_heappush_obj, 4, 4, mod_utimeq_heappush);
+
+STATIC mp_obj_t mod_utimeq_heappop(mp_obj_t heap_in, mp_obj_t list_ref) {
+    mp_obj_utimeq_t *heap = get_heap(heap_in);
+    if (heap->len == 0) {
+        nlr_raise(mp_obj_new_exception_msg(&mp_type_IndexError, "empty heap"));
+    }
+    mp_obj_list_t *ret = MP_OBJ_TO_PTR(list_ref);
+    if (!MP_OBJ_IS_TYPE(list_ref, &mp_type_list) || ret->len < 3) {
+        mp_raise_TypeError(NULL);
```
husigeza: NULL must be changed to a valid string because it makes the system crash in this form

PR #121, opened 2018-01-08
Add Bluetooth.CHAR_SUBSCRIBE_EVENT

semireg:
This has been built and tested on a WiPy 3.0 with latest firmware.

Design considerations:
• Naming of Bluetooth.CHAR_SUBSCRIBE_EVENT
• Naming of config()
• Exposing Bluetooth.CHAR_CONFIG_NOTIFY/INDICATE instead of simple bools on the characteristic
  
Forum discussion at https://forum.pycom.io/topic/2396/callback-when-ble-client-subscribes-to-notify-characteristic

### DIFF
--- a/esp32/mods/modbt.c
+++ b/esp32/mods/modbt.c
@@ -69,6 +69,7 @@
 #define MOD_BT_GATTS_WRITE_EVT                              (0x0010)
 #define MOD_BT_GATTC_NOTIFY_EVT                             (0x0020)
 #define MOD_BT_GATTC_INDICATE_EVT                           (0x0040)
+#define MOD_BT_GATTS_SUBSCRIBE_EVT                          (0x0080)
 
 /******************************************************************************
  DEFINE PRIVATE TYPES
@@ -197,6 +198,7 @@ typedef struct {
     uint32_t              events;
     uint32_t              trans_id;
     bool                  read_request;
+    uint16_t              config;
 } bt_gatts_char_obj_t;
 
 /******************************************************************************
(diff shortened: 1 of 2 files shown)


semireg:
Forum discussion at https://forum.pycom.io/topic/2396/callback-when-ble-client-subscribes-to-notify-characteristic

esp32/mods/modbt.c
```diff
@@ -566,6 +568,13 @@ static void gatts_event_handler(esp_gatts_cb_event_t event, esp_gatt_if_t gatts_
                 } else {    // descriptor
                     if (attr_obj->uuid.len == ESP_UUID_LEN_16 && attr_obj->uuid.uuid.uuid16 == GATT_UUID_CHAR_CLIENT_CONFIG) {
                         uint16_t value = param->write.value[1] << 8 | param->write.value[0];
+                        bt_gatts_char_obj_t *char_obj = (bt_gatts_char_obj_t *)attr_obj->parent;
+                        char_obj->config = value;
+                        char_obj->events |= MOD_BT_GATTS_SUBSCRIBE_EVT;
+                        if (char_obj->trigger & MOD_BT_GATTS_SUBSCRIBE_EVT) {
```
husigeza: Shouldn't this block be placed if the "value" checked and verified that it is the allowed range ? So the registered callback should be called/triggered if the configuration was successful which means the client successfully subscribed to the notify characteristic

semireg: There is definitely some ambiguity here that I didn't fully understand. The design I was trying to achieve generates a MOD_BT_GATTS_SUBSCRIBE_EVT callback for both SUBSCRIBE/UNSUBSCRIBE. That way, app can turn on/off resources on client demand. So this means whenever the descriptor is written to, MOD_BT_GATTS_SUBSCRIBE_EVT is called.

husigeza: Okay this makes sense if the aim was to trigger the callback when the client unsubscribes too.

PR #123, opened 2018-01-15
modbt: GATTCCharacteristic, add read_format()

semireg:
Add ability to read GATT descriptor Characteristic Presentation Format (0x2904). Returns descriptor value as bytes. Returns None when characteristic is missing the descriptor.

This is more a proof-of-concept because it currently lacks documentation and may not adhere to the standards of this project.

Cross-posted to forum at https://forum.pycom.io/topic/2460/ble-features-non-blocking-connects-char-presentation-format.

### DIFF
--- a/esp32/mods/modbt.c
+++ b/esp32/mods/modbt.c
@@ -129,6 +129,8 @@ typedef struct {
     uint32_t                events;
     uint16_t                value_len;
     uint8_t                 value[BT_CHAR_VALUE_SIZE_MAX];
+    uint16_t                format_len;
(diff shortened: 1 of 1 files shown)


husigeza:
Hello,
I reviewed and tested you implementation, it seems good.
I introduced a little optimization in bt_char_read_format() function.

husigeza:
Hello,

Development release 1.19.0.b2 contains an extended functionality with implementing the characteristic.read_descriptor(uuid) function which makes possible to read any descriptor based on the uuid.
Thanks for the idea and contribution!

PR #131, opened 2018-01-27
Simple font size scaling for framebuf

livius2:
I have implemented simple font size scaling as `framebuf_text_scaled`

usage from python size x 4
`framebuf.text_scaled('Ab', 0, 26, 4)`

also posted to official micropython as
https://github.com/micropython/micropython/pull/3583


livius2:
Same question here
I am curious, why it cannot be merged for so long time ;-)

PR #134, opened 2018-01-30
Endable a debug mode for modlte.c

brotherdust:
There are various useful debug messages that are commented out in modlte.c. I put each of them in a macro block (I think that's what it's called, correct me if I'm wrong) and un-commented. They can now be activated by setting LTE_DEBUG to 1

### DIFF
--- a/esp32/mods/modlte.c
+++ b/esp32/mods/modlte.c
@@ -83,6 +83,8 @@
 
 #define FILE_READ_SIZE 512
 
+#define LTE_DEBUG 1
+
 /******************************************************************************
  DECLARE PRIVATE DATA
  ******************************************************************************/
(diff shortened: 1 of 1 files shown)


doniks:
Hi @brotherdust it would actually be useful to have a reasonable debug mode! However, as of now (much to our own delay :( ) this conflicts a lot. Frankly it would need a rewrite. Including deciding a reasonable set of messages because many of these printf aren't even there anymore

I close this, since it needs a rewrite anyway. If you are still interested in this topic please make a new PR and feel free to ping me. 

Some nitpicking: 
- let's call it DEBUG_MODLTE 
- please use #ifdef instead of #if and do not indent the precompiler directives.

PR #139, opened 2018-02-17
ESP32: Add control for the Ctrl-F hot key

robert-hh:
The activity of the Ctrl-F key can be controlled like the Ctrl-C key
with micropython.kbd_intr().
kbd_intr(-1)
    Disable both Ctrl-C and Ctrl-F. This maintains backward compatibility
    to existing code.
kbd_intr(interrupt-char)
    Set the char for Interrupt by its ordinal value. The default would be 3 (Ctrl-C)
kbd_intr(interrupt-char, reset-char)
    Set the characters for both interrupt and reset. The typical
    values would be 3 (Ctrl-C) and 6 (Ctrl-F)

### DIFF
--- a/esp32/fatfs/src/drivers/sflash_diskio.c
+++ b/esp32/fatfs/src/drivers/sflash_diskio.c
@@ -43,7 +43,7 @@ static bool sflash_write (void) {
 DRESULT sflash_disk_init (void) {
     if (!sflash_init_done) {
         // this is how we diferentiate flash sizes in Pycom modules
-        if (esp_get_revision() > 0) {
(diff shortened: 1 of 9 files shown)


robert-hh:
@Xykon Hello Christian. I just noticed that you included in the development branch all changes but one, the change in modmicropython.c, which makes the PR accessible by a python script. Was that intentional or an omission?

robert-hh:
I just noticed that this PR is also incompletely included in the master branch.

robert-hh:
Almost fixed. Just telnet.c in the release-candidate branch is still due.

PR #140, opened 2018-02-21
CodingChallenge

iwahdan88:
Updated modos.c to add functinality to unix port for fdopen, fchmod, fclose and fchown


robert-hh:
To be consistent, the implementation of fdopen() must explicitely return a mp object of a certain type (MP_SMALL_INT), like the other fd... functions do. At the moment is does not.

PR #145, opened 2018-02-28
Allow RMT pulses to be sent asynchronously

Sympatron:
Added (optional) keyword argument 'wait_tx_done' to RMT.pulses_send() to
allow the function to return without haveing to wait for the
transmission to be done.
Also the GIL is unlocked during the transmission to allow other Python
code to execute.


husigeza:
Hello,

Thanks for the contribution, this PR has already been merged and the functionality is contained by development version 1.19.0.b1.

PR #146, opened 2018-02-28
Unlock the GIL while accessing queues

Sympatron:
Keeping the GIL locked while sending to full queues or receiving from
empty queues results in a deadlock, because no other Python code can
execute to put() or get() from the queue.


husigeza:
Hello,
Thanks for the contribution, the PR was merged into development release 1.19.0.b1.

PR #162, opened 2018-05-01
main.c, sflash_diskio.c: get flash size by call to spi_flash_get_chip_size()

robert-hh:
Instead of guessing it by ESP revision number. Even if the result from spi_flash_get_chip_size() could be
used directly here, the same logic is used in sflash_diskio.c and main.c to avoid inconsistencies.
The change in sflash_diskio.c was already implemented in previous versions but got lost or was dropped intentionally.

### DIFF
--- a/esp32/main.c
+++ b/esp32/main.c
@@ -52,6 +52,8 @@
 #include "mperror.h"
 #include "machtimer.h"
 
+#include "esp_spi_flash.h"
+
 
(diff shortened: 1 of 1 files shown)


robert-hh:
@Xykon There are problems with integration checks since a few days: continuous-integration/jenkins/pr-head

robert-hh:
Does not work as expected.

Issue #163, opened 2018-05-03
AU915 and US915 region code is too restrictive when adding 500kHz BW channels


Bdanilko:
System I'm using:
   (sysname='LoPy4', nodename='LoPy4', release='1.17.3.b1', version='ae3a294-dirty on 2018-05-01',  machine='LoPy4 with ESP32', lorawan='1.0.2', sigfox='1.0.1')
   I've done a new version of the firmware to test my 'fix' -- but the issue was definitely in the original 1.17.3.b1 firmware.

When adding a single 500kHz channel for the AU915 (id 64-71), it is not added:
`self.lora.add_channel(65, frequency=917500000,  dr_min=6,  dr_max=6)`

The issue seems to be that the function:
RegionAU915ChannelManualAdd( ChannelAddParams_t* channelAdd ) 
enforces that the dr_min has to be 0 for ALL channels. But for the AU915 and US915 channel id's 64-71, the dr_min is not meant to be zero (for AU915 it is DR6 and US915 it is DR4).

This results in the channel silently being discarded on the add_channel() call, so not being available for transmission.

smbunn:
What channels are predefined merley by using <lora = LoRa(mode=LoRa.LORAWAN, region=LoRa.EU868)>?

Bdanilko:
When a region is set, then all of the default channels are set for that region. Looking at the spec, that should be:
* 868.10 MHz (DR0-5)
* 868.30 MHz (DR0-5)
* 868.50 MHz (DR0-5)

New frequencies are added via the JoinAccept CFList or the NewChannelReq MAC command.

But the first three channels should always be there, as the two methods for adding channels can't change or remove the 3 defaults.

I haven't used EU868 yet, so this is all theoretical!

smbunn:
I agree, however this is clearly not working as when I used 868 I had no problems getting signals through to a LoRa gateway 7 km away but now I have specified AU915  nothing is getting through to me TTN gateway sitting 4 meters away.

Bdanilko:
Hi Simon,

Did you mean - region=LoRa.AU915 in your original post here? I commented on your thread in the pycom forums this morning that could be useful about the AU915 frequency plan.

There are some issues that make AU915 more difficult to get going. But with the latest firmware it is possible -- check the forum post. I'm running on AU915 as we speak.

Regards,
Brian

smbunn:
Yes my line of code should have read
```
[...]

PR #164, opened 2018-05-03
lib/lora: Allow non-zero dr_min for top channels in AU/US915 regions

Bdanilko:
The LoRaWAN-Regional-Parameters (1.1b) document, for AU915 and US915,
lists the channels 64 - 71 as having specific data rates which are not
zero. This change removes the check of dr_min == 0 for these channels.

Issue #163


Bdanilko:
Abandoning this pull request, and redoing it to see if we can get past the CLA issues.
https://forum.pycom.io/topic/3197/help-needed-signing-the-contributor-license-agreement/16

PR #183, opened 2018-07-14
lib/lora: Fix manual add of channels 64-71

Bdanilko:
The LoRaWAN-Regional-Parameters document (1.02 and 1.1.), for AU915 and US915,
lists the channels 64 - 71 as having specific data rates which are not
zero. This change removes the check of dr_min == 0 for these channels when
they are added manually.

Also a simple sanity check is done for the dr range.

This commit is in response to issue #163

### DIFF
--- a/lib/lora/mac/region/RegionAU915.c
+++ b/lib/lora/mac/region/RegionAU915.c
@@ -841,13 +841,19 @@ LoRaMacStatus_t RegionAU915ChannelManualAdd( ChannelAddParams_t* channelAdd )
         return LORAMAC_STATUS_PARAMETER_INVALID;
     }
 
-    // Validate the datarate range for min: must be DR_0
-    if( channelAdd->NewChannel->DrRange.Fields.Min != DR_0 )
+    // Validate the datarate range for min: must be DR_0 for channels 0-63
+    if( id < 64 && channelAdd->NewChannel->DrRange.Fields.Min != DR_0 )
     {
         drInvalid = true;
     }
     // Validate the datarate range for max: must be <= TX_MAX_DATARATE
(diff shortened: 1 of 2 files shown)


emmanuel-florent:
I have the same assumption you reported, gonna make more test and merge it soon :) thanks for the update.

Bdanilko:
> I have the same assumption you reported, gonna make more test and merge it soon :) thanks for the update.

Hi @emmanuel-florent -- I've been using the patch for months now with no problems.

emmanuel-florent:
Hi @Bdanilko you patch modify Semtech stack, Semtech source. Feel free to submit it on their repos but there is few chance it get accepted. For some reasons including certification we must stick to the official Semtech implementation thus it will not be merged for this reason.

Bdanilko:
Hi @emmanuel-florent 

Are you sure? I've looked at the semtech sources and those files don't have the functions (which I did NOT create but have slightly modified) for manually adding and removing channels. I've assumed that Pycom modified those files themselves, adding the new manual add/remove of channel functions that pycom needs-- you're git activity doesn't show that detail so I can't say for sure. Did one of your previous employees, or a contractor, modify the semtech sources instead of separating out that functionality? 

What is your plan going forward if you don't want to modify Pycom code in those files? -- the current situation is that there is a BUG in your code -- it doesn't match the documentation. At the very least, updates to the documentation are needed, listing the non-conformance.

The semtech stack doesn't have manual adds or deletes for the AU915 and US915 regions as those region plans allocate all of the channels. The reality, of course, is different as most gateways don't have 72 channel support (and the Pycom micro node only has 1 channel). Without the manual add/remove the Pycoms couldn't function with gateways that don't support all channels (all 72 of them).

PR #202, opened 2018-09-05
boards/SIPY/pins.csv: Fix wrong GPIO asssignments for P13 and P14

robert-hh:
Upon note of @prx, who also tested a firmware which was built according to this PR.
I faintly recall that this change had been addressed before in the past, but could not find any traces of that.


robert-hh:
Fixed in v1.1.20.rc0

PR #204, opened 2018-09-08
mods/modlte.c: Add option band=8 to the API lte.attach()

robert-hh:
That' all. Besides that, the function could be coded a little bit smaller, just iterating through a list of channel numbers, on the cost of RAM of course.


robert-hh:
Not sufficient to make band 8 accessible, and not adequate, if all bands will be available,

PR #226, opened 2018-11-14
Minor mod to build for custom python modules, and NVS string support

psychogenic:
Small change to allow for a standard, distinct, directory into which we can drop packages/modules to be frozen and included into the generated frozen_mpy.c

### DIFF
--- a/esp32/mods/modpycom.c
+++ b/esp32/mods/modpycom.c
@@ -9,6 +9,7 @@
 
 #include "py/mpconfig.h"
 #include "py/obj.h"
+#include "py/objstr.h"
 #include "py/runtime.h"
 #include "mperror.h"
 #include "updater.h"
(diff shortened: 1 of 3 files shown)


psychogenic:
Also add mods to pycom module to provide support for string set/get in NVS--now we have support strings and, with e.g. JSON, basically any data structure (up to the limit of 1984 bytes supported by the ESD-IDF nvs lib).

esp32/mods/modpycom.c
```diff
@@ -162,8 +163,41 @@ STATIC mp_obj_t mod_pycom_pulses_get (mp_obj_t gpio, mp_obj_t timeout) {
...
+
+
+/*
+ * nvs_setstring/nvs_getstring: support for NVS string storage and, 
+ * thanks to JSON/other serializers, pretty much any datastructure.
+ */
+STATIC mp_obj_t mod_pycom_nvs_setstring (mp_obj_t _key, mp_obj_t _value) {
+    const char *key = mp_obj_str_get_str(_key);
+    const char *value = mp_obj_str_get_str(_value);
+    if (strlen(value) >= 1984) {
```
husigeza: Where does 1984 come from ? As per [documentation](https://docs.espressif.com/projects/esp-idf/en/latest/api-reference/storage/nvs_flash.html) it is 5000.

esp32/mods/modpycom.c
```diff
@@ -162,8 +163,41 @@ STATIC mp_obj_t mod_pycom_pulses_get (mp_obj_t gpio, mp_obj_t timeout) {
...
+    }
+    esp_err_t esp_err = nvs_set_str(pycom_nvs_handle, key, value);
+    if (ESP_OK == esp_err) {
+        nvs_commit(pycom_nvs_handle);
+    } else if (ESP_ERR_NVS_NOT_ENOUGH_SPACE == esp_err || ESP_ERR_NVS_PAGE_FULL == esp_err || ESP_ERR_NVS_NO_FREE_PAGES == esp_err) {
+        nlr_raise(mp_obj_new_exception_msg(&mp_type_OSError, "no space available"));
+    } else if (ESP_ERR_NVS_INVALID_NAME == esp_err || ESP_ERR_NVS_KEY_TOO_LONG == esp_err) {
+        nlr_raise(mp_obj_new_exception_msg(&mp_type_ValueError, "invalid key (or too long)"));
+    } else {
+        // not ESP_OK, but not reporting?  TODO:check this...
```
husigeza: Please also raise an exception here with a generic exception message

esp32/mods/modpycom.c
```diff
@@ -178,17 +212,53 @@ STATIC mp_obj_t mod_pycom_nvs_set (mp_obj_t _key, mp_obj_t _value) {
...
+
+STATIC mp_obj_t mod_pycom_nvs_getstring (mp_obj_t _key) {
+    vstr_t vstr;
+    const char *key = mp_obj_str_get_str(_key);
+    size_t required_size = 0;
+    if (ESP_ERR_NVS_NOT_FOUND == nvs_get_str(pycom_nvs_handle, key, NULL, &required_size)) {
+        return mp_const_none;
+    }
+    vstr_init_len(&vstr, required_size);
+    if (ESP_OK != nvs_get_str(pycom_nvs_handle, key, vstr.buf, &required_size)) {
```
husigeza: Instead of this, after the necessary size is fetched I would simply allocate enough byte with m_malloc(), fetch the value from NVS, then create the return object with  returned string with mp_obj_new_str(), then free up the allocated memory and return the composed string. In this case it is not needed to compose vstr, remove closing \0 etc.

esp32/mods/modpycom.c
psychogenic: Well, this pull requests is getting pretty old...  
If you look at, say, https://docs.espressif.com/projects/esp-idf/en/v2.1.1/api-reference/storage/nvs_flash.html#keys-and-values

there's a note (in blue) "String and blob values are currently limited to 1984 bytes. For strings, this includes the null terminator."

husigeza: Yes, you are right, only in the newest versions (3.1, 3.2) it can be larger than 1984 bytes.

PR #234, opened 2018-12-13
Allow binding to LoRa mesh addresses

mgranberry:
The current LoRa mesh doesn't allow binding to an address, only the port.  Because of this it's impossible to send packets from the mesh-local EID, only the RLOC16-based addresses.  That's not ideal because the RLOC16-based address will change periodically.  This change allows binding to other addresses, potentially ones that don't change.


Xykon:
Sorry... I had some issues with the `v1.20.0.rc0` branch so I deleted it after creating the new release-candidate branch. However this also closed this PR so I restored it and re-opened the PR.

PR #235, opened 2018-12-15
utime: Revert back the ticks_xxx functions to use short ints.

robert-hh:
The ticks_xx functions wrap around at a certain value. For the initial micropython, this values is 2^30-1 (MP short int). For the pycom port, it is 2^32-1. Thats done by simply using C 32 bit integers. This implementation has two drawbacks:
-    uasyncio expects timers to be MP short ints.
-    using 32 bit ints cause the values to be standard ints in Python, which are more time-consuming 
     to deal with. That may hurt if for instance ticks_diff and ticks_xx are used a tight timing loop.
So this PR just goes back to the initial code, with was still present in the codebase and just had to be enabled with a config flag in mpconfigport.h.
Only ticks_cpu() is kept, limited to a 2^30 period, since different from the mp.org version is counts us ticks and the timing call is also used by the SX1272 and SX1276 driver. Side effects on the Lora nanogateway have to be tested yet, but I am pretty confident that it works. The ticks_cpu() call itself behaves well with ticks_diff().

### DIFF
--- a/esp32/hal/esp32_mphal.c
+++ b/esp32/hal/esp32_mphal.c
@@ -11,6 +11,7 @@
 #include <stdint.h>
 #include <stdbool.h>
 
+#include "py/smallint.h"
 #include "py/mpstate.h"
 #include "py/mphal.h"
 #include "py/runtime.h"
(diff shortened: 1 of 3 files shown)


Xykon:
Sorry... I had some issues with the `v1.20.0.rc0` branch so I deleted it after creating the new release-candidate branch. However this also closed this PR so I restored it and re-opened the PR.

PR #241, opened 2018-12-29
help: Simplify and extend help()

robert-hh:
Replace the port specific implementation for help() by the builtin version,
which merely consist of deleting all code but the specific help message.
At the same time, enable help('modules') to show a list of all builtin
and frozen module.

### DIFF
--- a/esp32/mods/machspi.c
+++ b/esp32/mods/machspi.c
@@ -195,8 +195,8 @@ STATIC void pybspi_transfer (mach_spi_obj_t *self, const char *txdata, char *rxd
     for (int i = 0; i < len; i += self->wlen) {
         uint32_t _rxdata = 0;
         uint32_t _txdata;
-        if (txdata) {
-            memcpy(&_txdata, &txdata[i], self->wlen);
+        if (txchar) {
+            _txdata = *txchar;
         } else {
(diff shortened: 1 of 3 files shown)


robert-hh:
Thanks. This is just reactivating what was already present. So all kudos go to the initial writers Damien, Paul and Daniel.
May I also direct you attention to another PR i made, also a reactivation of a previous mechanism by Paul (I guess). https://github.com/pycom/pycom-micropython-sigfox/pull/235
It makes the ticks_xx call cycle with a period of 2**30 and returns them as MP short int. In python scripts you will note a difference, but some modules like the asyncio use internally short int's for the time stamp, and these fail at them moment when use with the actual ticks_xx values. Not a critical issue, but when it fails it is hard to find. Asyncio fails after ~12 days.

robert-hh:
Sorry. It looked like a fix for spi.read() also joined this PR. It fixes.
spi.read(n, write=..)
The write parameter was ignored without that change.

robert-hh:
Thanks. Another open topic: The way ctrgl-f is handled is now fine is Master, but wrong in release-candidate. The latter was a  misunderstanding by @iwahdan88 about the intention. IMHO, the master branch implementation should be used

PR #253, opened 2019-01-25
build config: make openthread over lora a build-time configuration option

martijnthe:
Adds a make variable to disable openthread. By default it's still enabled.
To disable it, add `OPENTHREAD=off` to the `make` invocation.
This shaves off 160KB of the .bin, making the firmware fit again in the old OTA slot size of 1536KB.


iwahdan88:
Thanks @martijnthe

PR #255, opened 2019-02-01
Fixed delays causing not to achieve low power after deepsleep

arapostol:
esp32/lteppp.h:	added new machine state E_LTE_DEINIT
esp32/lteppp.c:	removed disabling PSM
		lteppp_deinit() changes lte state to E_LTE_DEINIT
		lteppp_deinit() removed sending setlpm command before disabling UART. This caused the state machine to wait for a response from modem and timeout, resulting in a delay.
esp32/modlte.c:
		removed disabling PSM
		removed sending setlpm command before lteppp_deinit(). Sending commands to modem before disabling UART cause delays.

### DIFF
--- a/esp32/lte/lteppp.c
+++ b/esp32/lte/lteppp.c
@@ -277,14 +277,10 @@ lte_legacy_t lteppp_get_legacy(void) {
 }
 
 void lteppp_deinit (void) {
-
-    // enable airplane low power mode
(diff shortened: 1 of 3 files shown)


amotl:
Dear @arapostol,

after looking at this PR, we speeded up LTE modem shutdown by simply not invoking the `detach` operation from the MicroPython API like `lte.deinit(detach=False, reset=True)` through [1].

Many thanks for the impulse!

With kind regards,
Andreas.

[1] https://github.com/hiveeyes/hiveeyes-micropython-firmware/commit/ec7c80e1219fb119c2881a0019cecdfad220f29f

PR #278, opened 2019-03-08
[WIP] Fix crash when pin ISR handler fires (i.e. LoRa TX done) during flash operation (i.e. ota_write() call)

martijnthe:
I was seeing crashes with "Cache disabled but cached memory region accessed" as reason.
The coredump showed one of the threads being inside `spi_flash_op_block_func()`, while the thread that caused the crash was an ISR (with LoRa's `OnRadioTxDone` handler somewhere in the stacktrace).
The comment above spi_flash_op_block_func(), suggests that if an interrupt handler accesses PSRAM, it must not be part of the "IRAM" ISR handlers, so it will be disabld during flash ops.
This change will flag all ISR handlers for the Pin module as non-IRAM and thereby preventing the crashes.

### DIFF
--- a/esp32/lora/sx1272-board.c
+++ b/esp32/lora/sx1272-board.c
@@ -14,6 +14,7 @@ Maintainer: Miguel Luis and Gregory Cristian
 */
 
 #include "board.h"
+#include "esp_attr.h"
 
 #if defined(LOPY) || defined (FIPY)
 
@@ -25,7 +26,7 @@ Maintainer: Miguel Luis and Gregory Cristian
 /*!
  * Radio driver structure initialization
  */
-const struct Radio_s Radio =
+DRAM_ATTR const struct Radio_s Radio =
 {
     SX1272Init,
     SX1272GetStatus,
--- a/esp32/lora/sx1276-board.c
+++ b/esp32/lora/sx1276-board.c
@@ -14,6 +14,7 @@ Maintainer: Miguel Luis and Gregory Cristian
 */
 
 #include "board.h"
+#include "esp_attr.h"
 
 #if defined(LOPY4)
 
@@ -24,7 +25,7 @@ Maintainer: Miguel Luis and Gregory Cristian
 /*!
  * Radio driver structure initialization
  */
-const struct Radio_s Radio =
+DRAM_ATTR const struct Radio_s Radio =
 {
     SX1276Init,
     SX1276GetStatus,
--- a/lib/lora/mac/region/Region.c
+++ b/lib/lora/mac/region/Region.c
@@ -629,7 +629,7 @@ bool RegionIsActive( LoRaMacRegion_t region )
     }
 }
 
-PhyParam_t RegionGetPhyParam( LoRaMacRegion_t region, GetPhyParams_t* getPhy )
+IRAM_ATTR PhyParam_t RegionGetPhyParam( LoRaMacRegion_t region, GetPhyParams_t* getPhy )
 {
     PhyParam_t phyParam = { 0 };
     switch( region )
--- a/lib/lora/mac/region/RegionEU868.c
+++ b/lib/lora/mac/region/RegionEU868.c
@@ -184,7 +184,7 @@ static uint8_t CountNbOfEnabledChannels( bool joined, uint8_t datarate, uint16_t
     return nbEnabledChannels;
 }
 
-PhyParam_t RegionEU868GetPhyParam( GetPhyParams_t* getPhy )
+IRAM_ATTR PhyParam_t RegionEU868GetPhyParam( GetPhyParams_t* getPhy )
 {
     PhyParam_t phyParam = { 0 };
 


martijnthe:
@Xykon @danicampora pretty nasty bug that I ran into this morning.

I don't have a repro snippet for you, but the way I was hitting often it involved:
- having LoRa repeatedly trying to join a non-existing/offline network
- use `ota_write()` with a relatively big buffer at the same time

iwahdan88:
@martijnthe I agree there is an issue here, indeed setting `non-IRAM` status to GPIO ISRs means that they will not be called when cache is disabled, however this might cause misses of GPIO interrupts when the cache is disabled. 
Can you share the stack trace so I can look more further into it.

martijnthe:
> Can you share the stack trace so I can look more further into it.

Sorry I've not saved the .elf (was using a one off custom build).

```
==================== CURRENT THREAD STACK =====================
#0  spi_flash_op_block_func (arg=0x0) at /Users/iwahdan/esp/pycom-esp-idf/components/spi_flash/cache_utils.c:82
#1  0x40081fae in ipc_task (arg=<optimized out>) at /Users/iwahdan/esp/pycom-esp-idf/components/esp32/ipc.c:67
```

esp32's coredump logic thought the crashed thread's TCB was not valid. It printed out some raw addresses though, which contained OnRadioTxDone.

iwahdan88:
Thanks @martijnthe, 
I want to ask if you were using the Flag `socket.SO_CONFIRMED` in setsockopt, 
I am asking because I had a look at the code and the definition of `OnRadioTxDone`,  I think I have an idea about what might be the root cause for this, in `OnRadioTxDone` The function `RegionGetPhyParam` is called when Class is set to C or Node ACK is requested, `RegionGetPhyParam` function definition does not have the `IRAM_ATTR` Macro in it, so compiler put this function in Flash instead of IRAM, thus when this ISR is called, a cache access is requested while the cache might be disabled by flash Ops.  did not test adding the 'IRAM_ATTR' to definition would fixes the problem yet, but maybe you want to test it .

martijnthe:
> I want to ask if you were using the Flag socket.SO_CONFIRMED in setsockopt,

Yes, we are using that.

> I am asking because I had a look at the code and the definition of OnRadioTxDone, I think I have an idea about what might be the root cause for this, in OnRadioTxDone The function RegionGetPhyParam is called when Class is set to C or Node ACK is requested, RegionGetPhyParam function definition does not have the IRAM_ATTR Macro in it, so compiler put this function in Flash instead of IRAM, thus when this ISR is called, a cache access is requested while the cache might be disabled by flash Ops. did not test adding the 'IRAM_ATTR' to definition would fixes the problem yet, but maybe you want to test it .

Aha, good find. If that's the issue, adding the `IRAM_ATTR` macro sounds like a better fix. I will try it out coming week.

iwahdan88:
@martijnthe great, Thanks!

martijnthe:
@iwahdan88 I just looked into adding `IRAM_ATTR`. But it seems quite involved since there are a lot of places where it would need to be added (every possible call path). I don't know of a quick way to do this automatically (there's probably a way leveraging the compiler or linker's knowledge about the code) and I'm afraid I will miss places when doing it manually.

So I'm going to leave it as-is for now (at the cost of potentially missing interrupts)...

iwahdan88:
> But it seems quite involved since there are a lot of places where it would need to be added (every possible call path)

Well I have checked the possible paths inside `onRadioTxDone` and seems that `IRAM_ATTR` is only missing for the definition of `RegionGetPhyParam` line 632 in `Region.c`, you can try with just adding `IRAM_ATTR` to only that function definition, for the paths inside that function there are currently none, this function actually always returns 0.

martijnthe:
Yeah but there are bunch of other ISRs. Shouldn't the code for those also be put into the IRAM section?

iwahdan88:
The other GPIO interrupts either have a C intr_handler or handled by the Micropython interrupt handler.
C intr_handlers are set by the call to function `machpin_register_irq_c_handler` in C, where the pointer to function passed should have the IRAM_ATTR , this is respected in the one place `machpin_register_irq_c_handler` is called from `GpioMcuSetInterrupt` in `lora/gpio-board.c` , as for Micropython interrupt handlers all functions servicing the Interrupts have `IRAM_ATTR` and the handling of interrupts is done in a FreeRTOS task `TASK_Interrupts`

martijnthe:
OK, I see that `SX1272OnDio0Irq` is the only place that is registered through `machpin_register_irq_c_handler` that ends up in a call to LoRaMac.c (`OnRadioTxDone`).
I guess that's what you are referring to?

What about the timer handlers? Glancing at lora/system/timer.c and esp/lora/timer-board.c, it looks like the timer handlers from LoRaMac.c would run in in interrupt context too or am I missing something?

iwahdan88:
> OK, I see that SX1272OnDio0Irq is the only place that is registered through machpin_register_irq_c_handler that ends up in a call to LoRaMac.c (OnRadioTxDone).
I guess that's what you are referring to?

yes

> What about the timer handlers? Glancing at lora/system/timer.c and esp/lora/timer-board.c, it looks like the timer handlers from LoRaMac.c would run in in interrupt context too or am I missing something?

Yes indeed these functions would be in Interrupt context, however ` Radio.Sleep( )`, `TimerStart`, `TimerSetValue`, `RegionSetBandTxDone` all have the `IRAM_ATTR` except for `RegionGetPhyParam` .
If you traced call tree from `OnRadioTxDone` you will find every function call is to an IRAM function except the call to `RegionGetPhyParam`

martijnthe:
What about `OnRadioRxDone`in LoRaMac.c? Shouldn't that one (and the functions called in there) also have `IRAM_ATTR`? It looks like it gets called from `SX1272RadioFlagsIrq` which is called from a timer ISR (`RadioIrqFlagsTimer`).

iwahdan88:
Well, Timer events here are not interrupts, timer callbacks are triggered deeply from RTOS task `prvTimerTask`, and this task runs on core 0 while OTA runs on Core 1 and both cores have separate cache for External Flash Access, so I Think that should not be a problem

martijnthe:
>  timer callbacks are triggered deeply from RTOS task prvTimerTask, and this task runs on core 0 while OTA runs on Core 1 and both cores have separate cache for External Flash Access, so I Think that should not be a problem

Ah, I missed that they were driven by a FreeRTOS timer. I was misled by the name `TimerIrqHandler` in lora/system/timer.c...

martijnthe:
@iwahdan88 ok, I'm a couple steps further. There were a couple more things that needed to be put into IRAM and the `Radio` global r/o variable also needed to be moved out of memory mapped flash and into DRAM because `Radio` was being accessed from an ISR (i.e. `Radio.Sleep()` in `onRadioTxDone`).

The good news: I'm not seeing crashes any more now.

The bad news is that I'm noticing RX'd packets are not received in my test scenario...
I'm repeatedly hitting the flash (writing to LittleFS) every ~1 second. LoRa join does not succeed when I do this. When I'm not hitting the flash it joins without problems. I noticed this line:
```
static IRAM_ATTR void SX1272OnDioIrq (void) {
...
irqflags1 = SX1272Read(REG_LR_IRQFLAGS);
...
```
Is it OK to use the SPI peripheral while a flash operation is happening?

Is there a way to log something over the UART from an ISR? Simply using `printf()` causes a crash.

martijnthe:
> Is there a way to log something over the UART from an ISR? Simply using printf() causes a crash.

I ended up copy & pasting the logging code from panic.c :)

I added a log to `SX1272OnDioIrq` to print out the result of `SX1272Read(REG_LR_IRQFLAGS)`.
I annotate the logs below with the #defines names of the flags:

Happy (no flash operation happening):
```
Joining LoRa...
F:00000008 => RFLR_IRQFLAGS_TXDONE
F:00000040 => RFLR_IRQFLAGS_RXDONE
SX1272RadioFlagsIrq:00000002
join accept: joined!
```

Sad (with flash operation happening):
```
Joining LoRa...
F:00000008 => RFLR_IRQFLAGS_TXDONE
F:00000080 => RFLR_IRQFLAGS_RXTIMEOUT
SX1272RadioFlagsIrq:00000001
F:00000080 => RFLR_IRQFLAGS_RXTIMEOUT
SX1272RadioFlagsIrq:00000001
```

So, it seems like the ISR is firing fine, but the problem is on the side of the LoRa radio chip... (really not receiving the message and timing out)

Could the flash operation be causing interference...?

martijnthe:
I crafted this minimal piece of code that consistently reproduces the problem.

```
from binascii import unhexlify
from network import LoRa
from socket import socket, AF_LORA, SOCK_RAW, SOL_LORA, SO_DR, SO_CONFIRMED
from pycom import wdt_on_boot
from time import sleep

wdt_on_boot(False)

lora = LoRa(mode=LoRa.LORAWAN, region=LoRa.EU868,
            tx_retries=1, power_mode=LoRa.TX_ONLY)

socket = socket(AF_LORA, SOCK_RAW)
socket.setsockopt(SOL_LORA, SO_DR, 5)
socket.setsockopt(SOL_LORA, SO_CONFIRMED, True)
socket.setblocking(False)

app_eui = unhexlify('REPLACE WITH YOUR APP EUI')
app_key = unhexlify('REPLACE WITH YOUR APP KEY')

while True:
    print('Joining LoRa...')
    lora.join(activation=LoRa.OTAA, auth=(app_eui, app_key), timeout=0)

    while not lora.has_joined():
        # sleep(1)  # sleeping instead of flash write: join will succeed
        with open('/flash/test.txt', 'wb') as f:
            f.write(b'blah')

    print('Joined LoRa successfully')
```

I added this logging to the firmware:
https://gist.github.com/martijnthe/712a92d1353ae01881dc17e73e0cbf28

Most of the time you will see:
```
Joining LoRa...
E_LORA_STATE_JOIN 0
F:00000008
F:00000080 ======> RX TIMEOUT!
SX1272RadioFlagsIrq:00000001
F:00000080
SX1272RadioFlagsIrq:00000001
MlmeConfirm status: 4 type: 0
```

If you comment out the code that writes to the file and uncomment the sleep, the LoRa join will succeed.

iwahdan88:
Thanks very much @martijnthe , I'll try to reproduce the issue at my side.

martijnthe:
@iwahdan88 thanks, looking forward to what you find. Note: you also need to apply this PR, otherwise you'll be crashing the firmware.

martijnthe:
@iwahdan88 I haven't checked it, but just had a theory:
You mentioned the lora timer port uses a FreeRTOS timer under the hood and timer callbacks don't need to be in IRAM. I noticed that the starting of the RX window is done by the lora software stack on the ESP32 (previously I had assumed the lora chip was in control over the starting of the window, but upon inspection it doesn't look like it). I noticed calls to `RxWindowSetup()` in `OnRxWindow1TimerEvent` and `OnRxWindow2TimerEvent`, both of which don't have the IRAM attr, so they won't be run during a flash op. Theory: the timer handlers are probably delayed and thus causes the RX window to be opened too late, thus missing the message from the gateway.

I don't have time to check this (let alone fix it) in the coming days. Hopefully you can take a look. Seems like a serious problem of the Lopy....

martijnthe:
Update;
OK, I checked my theory by logging something when the RxWindow1Timer is set and when it fires.
It should fire after exactly 4989ms, but using the bug repo code above (with repeated flash write), the timer is off by *a lot* – I've seen it fire like half a minute (!) later. I don't quite understand why it's off by so much though. The flash write is small so surely the FreeRTOS timer should be serviced much more often than that (although that makes me wonder whether a long flash operation could cause the FreeRTOS timer to become grossly inaccurate).

Digging further now to understand why the timer is only serviced so little when flash writes are happening.

iwahdan88:
@martijnthe Yes I think you’re correct, i realized that this evening, the RTOS timer seems impacted by the flash operations, I think the interrupts driving these timer ticks might be disabled during flash operations

martijnthe:
@iwahdan88 I'm a couple steps further again.
The lora timer port implementation (timer-board.c) is a bit weird, it's callback is called every 1ms and keeps it's own counter causing the time to drift...
The tick callback is structurally delayed when flash ops are happening. Therefore the total final delay is grossly delayed.

I see 2 solutions:

1. Directly use the FreeRTOS timer API instead of having the callback increment every 1ms (which aside of the drift issues, also seems wasteful to me). This will not solve the problem 100% (I've seen the 1ms tick callback be delayed by as much as ~3 seconds) but should be much more accurate than what it is now.
2. Use a hardware timer. I'm not sure if the HW timer TIMER_GROUP_1 is used (couldn't find a reference to that symbol in the code). This comes with the burden of having to put everything in IRAM.

Thoughts?

iwahdan88:
@martijnthe yes, I think hardware timers would be a good solution in that case , I need to get in touch with Espressif first to see if we can make the RTOS timers ticks not impacted by flash access, before going to that solution. Because as you mentioned that would cause extra memory in IRAM

iwahdan88 (approved):

iwahdan88:
@martijnthe , so we decided to go with Hardware timers, we will use timer groups for that, your pull request is still applicable so I will merge it .
Thanks very much @martijnthe  for your help in solving this issue.

martijnthe:
Sounds good to me. Go ahead and merge it.

Re fix with HW timer: do you have an ETA for when your team will fix this? Thanks much!

iwahdan88:
@martijnthe , I will try to include it in the next release-candidate release, around the beginning of next week

martijnthe:
@iwahdan88 unfortunately, I've managed to hit another path that causes a crash... (again "Cache disabled but cached memory region accessed" error).
The program counter points to 0x40084b54. Below is the disassembly around that address.
I'm not familiar with xtensa asm, I'm not sure what C code it maps to (the objdump output isn't really all that helpful there...). It looks like an program address is supposed to be constructed and then jump to it. `a8` contains `0xbad00bad`, suggesting the load instruction (0x40084b52) failed.

I noticed some other functions that can be called from `RegionEU868GetPhyParam` that don't have the IRAM attribute. I tried adding it, but it didn't help. There are also a few static variables that can potentially get used. Tried adding the DRAM attribute, but didn't help either...

For now I'm just going to create a work-around to prevent using LoRa + flash at the same time to avoid the crashes....

```
IRAM_ATTR PhyParam_t RegionEU868GetPhyParam( GetPhyParams_t* getPhy )
{
40084b3c:       004136          entry   a1, 32
    PhyParam_t phyParam = { 0 };

    switch( getPhy->Attribute )
40084b3f:       0288            l32i.n  a8, a2, 0
40084b41:       791c            movi.n  a9, 23
40084b43:       f7c882          addi    a8, a8, -9
40084b46:       02b987          bgeu    a9, a8, 40084b4c <RegionEU868GetPhyParam+0x10>
40084b49:       002ac6          j       40084bf8 <RegionEU868GetPhyParam+0xbc>
40084b4c:       efaa91          l32r    a9, 400809f4 <_init_end+0x5f4>
40084b4f:       a08890          addx4   a8, a8, a9
40084b52:       0888            l32i.n  a8, a8, 0
40084b54:       0008a0          jx      a8
40084b57:       a10000          sll     a0, a0
            phyParam.Value = MaxPayloadOfDatarateRepeaterEU868[getPhy->Datarate];
            break;
        }
        case PHY_DUTY_CYCLE:
        {
            phyParam.Value = EU868_DUTY_CYCLE_ENABLED;
```

amotl:
Dear @martijnthe,

> For now I'm just going to create a work-around to prevent using LoRa + flash at the same time to avoid the crashes....

> Unfortunately, I've managed to hit another path that causes a crash... (again "Cache disabled but cached memory region accessed" error).

We have been working on mitigating these issues, so if you feel lucky, you might want to try our custom build just released on [1]. More background about this is available through [2]. We will be happy to get any feedback if this would work for you without applying any workarounds.

If we are lucky together, this will improve the stability significantly. If you will be still receiving the core dumps, I will be happy if you would share its content with us.

Please be aware that when upgrading to the current 1.20.1 development release these builds are based upon, you will have to erase your device completely before flashing in order to keep things straight. You will find respective references to this on the [Pycom user forum](https://forum.pycom.io/) and on [3]. Hint: Use a command like
```
pycom-fwtool-cli --port /dev/ttyUSB0 erase_all
```

While I believe you might know your way around this already, I still would like to mention that: Please also be aware that this procedure will also erase the LoRa MAC stored on the device. @robert-hh thankfully outlined the procedure to save and/or restore it appropriately:

> - If you still know the value the LoraMac had before, then you can follow the procedure explained at https://forum.pycom.io/topic/1272/lora-mac-ffffffffffffffff/21.
> - If you do not know it, you will have to use the Pycom updater to restore it:
>   - First, use the Pycom updater (GUI version) to install a recent image (like pybytes 1.20.1) from the Internet. You do not have to use it, the installation process itself already restores both the LoRa MAC and the Sigfox ID.
>   - Then, use the Pycom updater to flash the custom image. Do not erase the flash in between.

As it turned out to gain more robustness for others already [4,5], we will be happy to learn if this happens to you as well.

With kind regards,
Andreas.

[1] https://packages.hiveeyes.org/hiveeyes/foss/pycom/vanilla/LoPy4-1.20.1.r1-0.6.0-vanilla-dragonfly.tar.gz
[2] https://community.hiveeyes.org/t/investigating-random-core-panics-on-pycom-esp32-devices/2480
[3] https://community.hiveeyes.org/t/installing-the-recent-pycom-firmware-1-20-1-r1-requires-erasing-the-flash-memory-completely/2688
[4] https://community.hiveeyes.org/t/testing-the-custom-dragonfly-builds-on-pycom-devices/2746
[5] https://github.com/pycom/pycom-micropython-sigfox/issues/361#issuecomment-553399627

amotl:
Dear @martijnthe,

we appreciate that you liked what we just posted here. May I humbly ask if you had the chance to test this on any real iron already? We are still looking around for people capable enough testing it within real scenarios using different bearers and especially when accessing the filesystem from within MicroPython to give us more solid feedback about what we did here in order to gain more confidence about it. 

Thanks already and with kind regards,
Andreas.

martijnthe:
@amotl I’m not working on the project any more!
Good luck!

amotl:
Ok. Thanks for letting me know!

PR #336, opened 2019-09-02
PYFW-272: BLE crash after deinit

oligauc:
Bluetooth.advertise fails when Bluetooth object is re-initialized after bel.deinit()


oligauc:
Pull request to release-candidate not master

PR #346, opened 2019-10-07
mods/modlte.c: Fix lenght field for AT commands

robert-hh:
At three places the length field was not set when AT commands werde
issued, causing lte.attach() with the band=xx option to fail.
Also, affected: lte_get_modem_version()

### DIFF
--- a/esp32/mods/modlte.c
+++ b/esp32/mods/modlte.c
@@ -301,13 +301,10 @@ static bool lte_check_legacy_version(void) {
 
 static int lte_get_modem_version(void)
 {
-    lte_task_cmd_data_t cmd = { .timeout = LTE_RX_TIMEOUT_MAX_MS };
-
     /* Get modem version */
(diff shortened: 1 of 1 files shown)


iwahdan88:
Thanks @robert-hh , i see you updated submodule axtls, is that intended?

robert-hh:
No. It's not intended. It seems to happen as part of the build process. I had troubles updating the submodules. But now it worked.

iwahdan88:
can you try `git submodule sync` and `git submodule update` as these modules where updated

robert-hh:
So I made another PR for the adc.init() case, based on the Dev branch. This time, only the file pybadc.c is shown as changed.

robert-hh:
OK. I cancel this one and give it another try.

robert-hh:
OK. A new PR. This time only one file is flagged as changed

PR #352, opened 2019-10-23
feat: Pybytes library release 1.2.0

jirkadev:
### Added
* firmware can be updated when deploying a release (from Pybytes)
* network settings can be changed when deploying a release (from Pybytes)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,7 @@
 *.dis
 *.exe
 
-# Packages 
+# Packages
 ############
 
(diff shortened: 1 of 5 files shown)


iwahdan88 (approved):

robert-hh:
Hello @iwahdan88 , the 1.20.1r1 versions are still not available on the firmware download page and will not be installed by the Pycom firmware updater. Is that an omission or intentional? if intentional, what are the alternatives.

PR #360, opened 2019-11-11
Implemented configuration options for BLE security

DvdGiessen:
For our product we needed the ability to configure some additional options related to BLE security, which were available in the lower level Espressif stack but were not exposed though any MicroPython interface.

This pull request exposes a few new constants, parameters and methods for configuring these properties. Namely:

* A new method, `bluetooth.set_pin(new_pin)`, which can be used to configure a new pin at any time, not just during initialization of the Bluetooth object.
* A few constants `Bluetooth.PERM_*`, which can be used with the pre-existing `permissions` parameter of the `.characteristic()` method to configure the permissions that will be applied to a newly created characteristic. A secure connection will be established as soon as an encrypted characteristic is touched, meaning that by setting permissions it is possible to specify unencrypted characteristics which may be accessed without pairing. By default the permissions are set to encrypted if security is enabled, this was not changed and thus everything will function exactly the same as before unless a developer chooses to specifically set the parameter.
* Two new named parameters for the `Bluetooth` initialization, namely `privacy=True` and `secure_connections=True`,  which allow selectively disabling certain security features which impact how phones and other devices behave when interacting with the Pycom board. Again, by default these remain enabled and unless changed explicitly everything will function the same as before.
* The existing `bluetooth.set_advertisement_params()` method would not work when BLE security was enabled. This has been resolved so that now it is possible to configure advertisement parameters as expected in all cases.
* The Python interface would not allow pin codes to start with a `0` digit, while it is supported and allowed by the Bluetooth spec. In all our BLE products the default pin code is `000000`, and for uniformity we want to use that here as well. This pull request resolves this so now pin codes are allowed to start with a zero.

The changes were verified with a number of phones, and we also double-checked by inspecting the raw BLE traffic with a sniffer.

Documentation has been updated to describe the new `set_pin()` method and permission constants.

### DIFF
--- a/esp32/mods/modbt.c
+++ b/esp32/mods/modbt.c
@@ -80,7 +80,6 @@
 #define MOD_BT_GATTS_CLOSE_EVT                              (0x0400)
 #define MOD_BT_NVS_NAMESPACE                                "BT_NVS"
 #define MOD_BT_HASH_SIZE                                    (20)
-#define MOD_BT_PIN_LENGTH                                   (6)
 
 /******************************************************************************
  DEFINE PRIVATE TYPES
@@ -103,6 +102,7 @@ typedef struct {
     bool                  advertising;
     bool                  controller_active;
     bool                  secure;
+    bool                  privacy;
 } bt_obj_t;
 
 typedef struct {
@@ -246,16 +246,6 @@ static esp_ble_adv_params_t bt_adv_params = {
     .channel_map        = ADV_CHNL_ALL,
     .adv_filter_policy  = ADV_FILTER_ALLOW_SCAN_ANY_CON_ANY,
 };
-
-static esp_ble_adv_params_t bt_adv_params_sec = {
-    .adv_int_min        = 0x100,
-    .adv_int_max        = 0x100,
-    .adv_type           = ADV_TYPE_IND,
-    .own_addr_type      = BLE_ADDR_TYPE_RANDOM,
-    .channel_map        = ADV_CHNL_ALL,
-    .adv_filter_policy  = ADV_FILTER_ALLOW_SCAN_ANY_CON_ANY,
-};
-
 
 static bool mod_bt_allow_resume_deinit;
 static uint16_t mod_bt_gatts_mtu_restore = 0;
@@ -416,11 +406,7 @@ void bt_resume(bool reconnect)
 
             /* See if there was an averstisment active before Sleep */
             if(bt_obj.advertising) {
-                if (!bt_obj.secure){
-                    esp_ble_gap_start_advertising(&bt_adv_params);
-                } else {
-                    esp_ble_gap_start_advertising(&bt_adv_params_sec);
-                }
+                esp_ble_gap_start_advertising(&bt_adv_params);
             }
         }
 
@@ -472,22 +458,6 @@ static void create_hash(uint32_t pin, uint8_t *h_value)
     mbedtls_sha1_free(&sha1_context);
 }
 
-static bool is_pin_valid(uint32_t pin)
-{
-    int digits = 0;
-
-    while(pin != 0)
-    {
-        digits++;
-        pin /= 10;
-    }
- 
-    if (digits != MOD_BT_PIN_LENGTH){
-       return false;
-    }
-
-    return true;
-}
 static bool pin_changed(uint32_t new_pin) 
 {   
      bool ret = false;
@@ -517,29 +487,42 @@ static bool pin_changed(uint32_t new_pin)
      return ret;
 }
 
-static void remove_all_bonded_devices(void)
+static void remove_bonded_devices(void)
 {
     int dev_num = esp_ble_get_bond_device_num();
 
     esp_ble_bond_dev_t *dev_list = heap_caps_malloc(sizeof(esp_ble_bond_dev_t) * dev_num, MALLOC_CAP_INTERNAL | MALLOC_CAP_8BIT);
     esp_ble_get_bond_device_list(&dev_num, dev_list);
     for (int i = 0; i < dev_num; i++) {
+        if (bt_obj.gatts_conn_id >= 0 && memcmp((void *) dev_list[i].bd_addr, (void *) bt_obj.client_bda, sizeof(esp_bd_addr_t)) == 0) {
+            continue;
+        }
         esp_ble_remove_bond_device(dev_list[i].bd_addr);
     }
 
     free(dev_list);
 }
 
-static void set_secure_parameters(uint32_t passKey){  
-
-    if (pin_changed(passKey)) {
-       remove_all_bonded_devices();
+static void set_pin(uint32_t new_pin)
+{
+    if (!bt_obj.secure) {
+        nlr_raise(mp_obj_new_exception_msg(&mp_type_ValueError, "Cannot set PIN when secure is not set"));
     }
-  
-    uint32_t passkey = passKey;
-    esp_ble_gap_set_security_param(ESP_BLE_SM_SET_STATIC_PASSKEY, &passkey, sizeof(uint32_t)); 
-    
-    esp_ble_auth_req_t auth_req = ESP_LE_AUTH_REQ_SC_MITM_BOND;
+
+    if (new_pin > 999999) {
+        nlr_raise(mp_obj_new_exception_msg(&mp_type_ValueError, "Only 6 digit (0-9) pins are allowed"));
+    }
+
+    if (pin_changed(new_pin)) {
+       remove_bonded_devices();
+    }
+
+    uint32_t passkey = new_pin;
+    esp_ble_gap_set_security_param(ESP_BLE_SM_SET_STATIC_PASSKEY, &passkey, sizeof(uint32_t));
+}
+
+static void set_secure_parameters(bool secure_connections) {
+    esp_ble_auth_req_t auth_req = secure_connections ? ESP_LE_AUTH_REQ_SC_MITM_BOND : ESP_LE_AUTH_BOND;
     esp_ble_gap_set_security_param(ESP_BLE_SM_AUTHEN_REQ_MODE, &auth_req, sizeof(uint8_t));
 
     esp_ble_io_cap_t iocap = ESP_IO_CAP_OUT;
@@ -640,11 +623,6 @@ static void gap_events_handler (esp_gap_ble_cb_event_t event, esp_ble_gap_cb_par
         xQueueSend(xGattsQueue, (void *)&gatts_event, (TickType_t)0);
         break;
     }
-    case ESP_GAP_BLE_NC_REQ_EVT:
-    case ESP_GAP_BLE_PASSKEY_NOTIF_EVT: {
-         printf("BLE paring passkey : %d\n", param->ble_security.key_notif.passkey);
-         break;
-    }
     case ESP_GAP_BLE_SEC_REQ_EVT: {
         if (bt_obj.secure){
             esp_ble_gap_security_rsp(param->ble_security.ble_req.bd_addr, true);
@@ -691,7 +669,7 @@ static void gattc_events_handler(esp_gattc_cb_event_t event, esp_gatt_if_t gattc
         status = p_data->reg.status;
         bt_obj.gattc_if = gattc_if;
         if (bt_obj.secure){
-           esp_ble_gap_config_local_privacy(true);
+           esp_ble_gap_config_local_privacy(bt_obj.privacy);
         }
         break;
     case ESP_GATTC_OPEN_EVT:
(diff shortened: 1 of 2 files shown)


esp32/mods/modbt.c
```diff
@@ -1445,35 +1415,47 @@ static mp_obj_t modbt_connect(mp_obj_t addr)
 
 STATIC mp_obj_t bt_set_advertisement_params (mp_uint_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_args) {
```
knagymate, line 1416: The problem with original code is that if we call `bt_set_advertisement_params()` with keyword argument(s), than unmentioned `bt_adv_params` will be reinitiated, which is probably not the intended behaviour of the function.

Example with **original** code:
1. Calling `bt_set_advertisement_params(adv_int_max = 128)` 
      - `bt_adv_params.adv_int_min = 32`
      - `bt_adv_params.adv_int_max = 128`
1. Calling `bt_set_advertisement_params(adv_int_min = 16)` 
      - `bt_adv_params.adv_int_min = 16`
      - **`bt_adv_params.adv_int_max = 64`** <- adv_int_max is reinitialized

Example with **new** code:
1. Calling `bt_set_advertisement_params(adv_int_max = 128)` 
      - `bt_adv_params.adv_int_min = 32`
      - `bt_adv_params.adv_int_max = 128`
1. Calling `bt_set_advertisement_params(adv_int_min = 16)` 
      - `bt_adv_params.adv_int_min = 16`
      - **`bt_adv_params.adv_int_max = 128`** <- adv_int_max remains the same

The behaviour is tested on LoPy4, I suggest to accept all the modifications inside `bt_set_advertisement_params()` function definition.

esp32/mods/modbt.c
```diff
@@ -1075,24 +1040,29 @@ static mp_obj_t bt_init_helper(bt_obj_t *self, const mp_arg_val_t *args) {
...
-    { MP_QSTR_modem_sleep,  MP_ARG_KW_ONLY  | MP_ARG_OBJ,   {.u_obj  = MP_OBJ_NULL} },
-    { MP_QSTR_secure,       MP_ARG_KW_ONLY  | MP_ARG_BOOL,  {.u_bool = false} },
-    { MP_QSTR_pin,          MP_ARG_KW_ONLY  | MP_ARG_INT,   {.u_int  = 123456} },
-    { MP_QSTR_mtu,          MP_ARG_KW_ONLY  | MP_ARG_INT,   {.u_int  = BT_MTU_SIZE_MAX} },
+    { MP_QSTR_id,                                     MP_ARG_INT,   {.u_int  = 0} },
+    { MP_QSTR_mode,                 MP_ARG_KW_ONLY  | MP_ARG_INT,   {.u_int  = E_BT_STACK_MODE_BLE} },
+    { MP_QSTR_antenna,              MP_ARG_KW_ONLY  | MP_ARG_OBJ,   {.u_obj  = MP_OBJ_NULL} },
+    { MP_QSTR_modem_sleep,          MP_ARG_KW_ONLY  | MP_ARG_OBJ,   {.u_obj  = MP_OBJ_NULL} },
+    { MP_QSTR_secure,               MP_ARG_KW_ONLY  | MP_ARG_BOOL,  {.u_bool = false} },
+    { MP_QSTR_pin,                  MP_ARG_KW_ONLY  | MP_ARG_INT,   {.u_int  = 123456} },
```
knagymate, line 1062: No change.

esp32/mods/modbt.c
```diff
@@ -1075,24 +1040,29 @@ static mp_obj_t bt_init_helper(bt_obj_t *self, const mp_arg_val_t *args) {
...
+    { MP_QSTR_privacy,              MP_ARG_KW_ONLY  | MP_ARG_BOOL,  {.u_bool = true} },
+    { MP_QSTR_secure_connections,   MP_ARG_KW_ONLY  | MP_ARG_BOOL,  {.u_bool = true} },
+    { MP_QSTR_mtu,                  MP_ARG_KW_ONLY  | MP_ARG_INT,   {.u_int  = BT_MTU_SIZE_MAX} },
```
knagymate, line 1065: No change.

esp32/mods/modbt.c
```diff
@@ -517,29 +487,42 @@ static bool pin_changed(uint32_t new_pin)
...
+    if (pin_changed(new_pin)) {
+       remove_bonded_devices();
+    }
+
+    uint32_t passkey = new_pin;
+    esp_ble_gap_set_security_param(ESP_BLE_SM_SET_STATIC_PASSKEY, &passkey, sizeof(uint32_t));
+}
+
+static void set_secure_parameters(bool secure_connections) {
+    esp_ble_auth_req_t auth_req = secure_connections ? ESP_LE_AUTH_REQ_SC_MITM_BOND : ESP_LE_AUTH_BOND;
```
knagymate, line 525: Further review needed.

esp32/mods/modbt.c
DvdGiessen: Correct, only indentation was changed to maintain alignment.

esp32/mods/modbt.c
DvdGiessen: Correct, only indentation was changed to maintain alignment.

esp32/mods/modbt.c
```diff
@@ -640,11 +623,6 @@ static void gap_events_handler (esp_gap_ble_cb_event_t event, esp_ble_gap_cb_par
         xQueueSend(xGattsQueue, (void *)&gatts_event, (TickType_t)0);
         break;
     }
-    case ESP_GAP_BLE_NC_REQ_EVT:
-    case ESP_GAP_BLE_PASSKEY_NOTIF_EVT: {
-         printf("BLE paring passkey : %d\n", param->ble_security.key_notif.passkey);
-         break;
-    }
```
knagymate, line 647: Do we want to inform the user about the `BLE paring passkey` on REPL, or is it just a remaining debug signal?

esp32/mods/modbt.c
```diff
@@ -472,22 +458,6 @@ static void create_hash(uint32_t pin, uint8_t *h_value)
...
-        digits++;
-        pin /= 10;
-    }
- 
-    if (digits != MOD_BT_PIN_LENGTH){
-       return false;
-    }
-
-    return true;
-}
```
knagymate, line 490: It is a problem that length of `pin` is not inspected anymore. If pin can start with `0`, and we store it in an integer, we cannot check the length. I have tested the commit and expreienced the following:

Initiated instance with `bluetooth = Bluetooth(secure=True, pin=1234)` on LoPy4, without getting any error (intended because of the deleted code snippet), and tried to bond from a phone, using nRF Connect. Using `1234` pin code, connection cannot be created, instead, with `001234`, it was OK.

There was no other issue with any 6 digit pin codes (i.e. `050399` or `000001`), and pin longer than 6 digits still gives an error (i.e. `1234567`).

Do we want to:

1. Ignore pins starting with zero (official version)
2. Handle pins starting with zeros, but doesn't check pin minimal length? (this commit)
3. Handle both length, and enable pin starting with zero? (new implementation, I think we cannot check length, as we represent pin as int -> `001` == `000001`).

DvdGiessen: This is intentional and as expected. As a direct result of the BLE implementation using an integer type to represent the passkey the leading zero's may be omitted in the calling Python code, but the resulting passkey will always be 6 digits. From the Bluetooth Core specification:
> [...] the Bluetooth passkey is a 6-digit numerical value. It is represented as integer value in the range 0x00000000 – 0x000F423F (000000 to 999999).

Note that this is also documented in the additional documentation for the user-facing `set_pin()` method.

esp32/mods/modbt.c
```diff
@@ -905,12 +883,6 @@ static void gatts_event_handler(esp_gatts_cb_event_t event, esp_gatt_if_t gatts_
                         if (char_obj->trigger & MOD_BT_GATTS_SUBSCRIBE_EVT) {
                             mp_irq_queue_interrupt(gatts_char_callback_handler, char_obj);
                         }
-
-                        if (value == 0x0001) {  // notifications enabled
-                            bt_gatts_char_obj_t *char_obj = (bt_gatts_char_obj_t *)attr_obj->parent;
-                            // the size of value[] needs to be less than MTU size
-                            esp_ble_gatts_send_indicate(gatts_if, param->write.conn_id, char_obj->attr_obj.handle, char_obj->attr_obj.value_len, char_obj->attr_obj.value, false);
-                        }
```
knagymate, line 913: I don't know why it is removed.

esp32/mods/modbt.c
```diff
@@ -963,20 +935,13 @@ static void gatts_event_handler(esp_gatts_cb_event_t event, esp_gatt_if_t gatts_
             if (bt_obj.trigger & MOD_BT_GATTS_CONN_EVT) {
                 mp_irq_queue_interrupt(bluetooth_callback_handler, (void *)&bt_obj);
             }
-            if (bt_obj.secure){
-                esp_ble_set_encryption(p->connect.remote_bda, ESP_BLE_SEC_ENCRYPT_MITM);
-            }
```
knagymate, line 968: It is okay anyways, `esp_ble_set_encryption(p->connect.remote_bda, ESP_BLE_SEC_ENCRYPT_MITM)` instruction causes that a bonding request is sent to master, when slave tries to just connect. As @DvdGiessen mentioned:

> A secure connection will be established as soon as an encrypted characteristic is touched, meaning that by setting permissions it is possible to specify unencrypted characteristics which may be accessed without pairing. By default the permissions are set to encrypted if security is enabled, this was not changed and thus everything will function exactly the same.

This modification is OK.

esp32/mods/modbt.c
```diff
@@ -517,29 +487,42 @@ static bool pin_changed(uint32_t new_pin)
...
+static void remove_bonded_devices(void)
 {
     int dev_num = esp_ble_get_bond_device_num();
 
     esp_ble_bond_dev_t *dev_list = heap_caps_malloc(sizeof(esp_ble_bond_dev_t) * dev_num, MALLOC_CAP_INTERNAL | MALLOC_CAP_8BIT);
     esp_ble_get_bond_device_list(&dev_num, dev_list);
     for (int i = 0; i < dev_num; i++) {
+        if (bt_obj.gatts_conn_id >= 0 && memcmp((void *) dev_list[i].bd_addr, (void *) bt_obj.client_bda, sizeof(esp_bd_addr_t)) == 0) {
+            continue;
+        }
```
knagymate, line 499: If a see correctly, the aim of this part is to not remove bond with device which is currently connected.

I tested this part, and bond won't be removed between LoPy4 and connected peripheral, but as you disconnect, can't be reconnect and rebond to LoPy4 anymore, I could not seen its advertisement anymore.

esp32/mods/modbt.c
husigeza: These lines should be kept to send notifications in case of WRITE events when enabled.

esp32/mods/modbt.c
DvdGiessen: This was tested and confirmed to work as expected on a LoPy4, FiPy and GPy, and we are seeing advertisements in the raw traffic. The sending of advertisements is not handled by this part of the code, so I'm not sure what causes the issue in your test but it seems like it wouldn't be related to the changes in these lines.

esp32/mods/modbt.c
DvdGiessen: The changes to these lines are for of the new `secure_connections` parameter described in the pull request's description.

esp32/mods/modbt.c
DvdGiessen: Since the changes in this pull request makes the passkey configurable from MicroPython, there was no need for this debugging print statement to remain. The updated documentation explains how a Bluetooth PIN may be set.

esp32/mods/modbt.c
knagymate: Yes, I have also tested, and I am sure that the problem is not coming from your changes, It is also an issue without it. I have tested with `LoPy4` and `nRF Connect` on my Android, and it both cases, I experience that:

1. Connect to LoPY4
2. Modify secured characteristics
3. Bonding devices
4. Set new pin
5. Disconnect (with your change) `or` Automatically disconnected (withhout the change)
6. Advertisement disappears (with manual `bluetooth.advertise(True)` it can be seen again)

So I recommended to add `esp_ble_gap_start_advertising(&bt_adv_params);` at the end of `remove_bonded_devices()` function.

esp32/mods/modbt.c
knagymate: Yes, thank you for improvement, we will add `secure_connections`.

husigeza:
Thanks for the contribution, we checked this PR and approved, it will be part of the next releases.
We introduced some minor changes:
- added secure_connections to bt_obj_t struct
- restored remove_all_bonded_devices() instad of remove_bonded_devices
- secure parameters are set inside set_pin()
- secure connection is decided based on PIN set
- bt_init_args secure argument is removed
- removed 'BLE paring passkey: ' REPL message
- removed automatic secure connection in case of connection
- restored characteristics notification

PR #377, opened 2019-12-12
Release xykon

Xykon:
Add mDNS support

Disable the LTE radio before resetting the modem


catalinio (approved):
All seems fine, probably @husigeza should take a look, too.

husigeza (approved):
Looks good for me !

PR #382, opened 2020-01-16
Fixing crash when make_new is called with zero arguments

knagymate:
When class is instantiated on python side with zero arguments `uart = UART(0)`, crash is experienced. Crash is caused by `mp_obj_get_int()` function on `MP_OBJ_NULL`. If there is no strong reason using `MP_ARG_OBJ` as default value of ID, it seems reasonable to use `MP_ARG_INT` instead, with the intended `MACH_UART_0` default value.

In case of keeping the original MP_ARG_OBJ definition of id, an other workaround could be:

`uint uart_id;`
`if(args[0].u_obj == MP_OBJ_NULL) {`
`uart_id = MACH_UART_0;`
`} else {`
`uart_id = mp_obj_get_int(args[0].u_obj);`
`}`

Initiating UART with `id = 0` (which is the default value) results a 'freeze-like' behaviour in REPL (we 'lost connection' to the board from terminal). In frozen module, terminal is started during boot on `id = 0`, `baudrate = 115200`, so this is the intended behaviour, but default UART id could be also `id = 1` (just a tip).

### DIFF
--- a/esp32/mods/machuart.c
+++ b/esp32/mods/machuart.c
@@ -451,7 +451,7 @@ error:
 }
 
 STATIC const mp_arg_t mach_uart_init_args[] = {
-    { MP_QSTR_id,                              MP_ARG_OBJ,  {.u_obj = MP_OBJ_NULL} },
+    { MP_QSTR_id,                              MP_ARG_INT,  {.u_int = MACH_UART_0} },
     { MP_QSTR_baudrate,                        MP_ARG_INT,  {.u_int = 9600} },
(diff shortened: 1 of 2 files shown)


husigeza:
@knagymate : Let me close this PR as somehow this gets mixed with another one. Please re-open this PR where the only change is related to UART, skipping the BLE modifications.

husigeza:
Closing temporary, new PR to be opened only with UART changes.

PR #390, opened 2020-01-28
Added interface to save and resume SSL sessions

DvdGiessen:
In many parts of the world, mobile data is both expensive and slow, and thus we are interested in reducing our mobile data use. One area where we might improve is in our use of TLS. Our product uses TLS using the built-in `ussl` module, and the overhead of TLS on these connections is quite considerable.

Performing a full TLS handshake is expensive, both in time (two round-trips) and in data (full certificates are transmitted). The first time we connect to a server there is not really a way around this. However for subsequent connections we can instead perform a TLS resumption, reducing both the number of round-trips and amount of data transferred.

TLS resumption is supported by the mbedtls library, but had no MicroPython interface exposed to actually use it. This pull requests implements a new function, `ussl.save_session(sock)`, which can be used to make a copy of the session data of an already established TLS connection. The object which it returns holds the necessary data to resume a TLS connection later on, independent of the SSLSocket (which may be closed and fully disposed). To use it a new optional parameter is added to the `ussl.wrap_socket()` function which takes a saved session object and uses it to resume a TLS connection.

(Note: We opted for manual saving an session object as opposed to an automatic caching mechanism. Our approach gives the user more control over which sessions are persisted and doesn't require a complexer interface to manage the cache.)

A quick test shows vast improvements in both time (from 0.8266 to 0.0987s ≈ 88% faster) and data use (upload from 1078 bytes to 1069 bytes, download from 4001 to 634 bytes, total ≈ 66% smaller) for resumed sessions (averages, including TCP overhead).

Unless explicitly used, these additions do not impact any existing code and are fully backwards compatible. Documentation has been updated with an example of the new functionality.

### DIFF
--- a/docs/library/ussl.rst
+++ b/docs/library/ussl.rst
@@ -114,6 +114,21 @@ network sockets, both client-side and server-side.
 
        SSL sockets inherit all methods and from the standard sockets, see the :mod:`usocket` module.
 
+    .. function:: ssl.save_session(sock)
+
+       Takes an instance sock of ssl.SSLSocket, and returns an instance of ssl.SSLSession representing saved session data from the socket, which can be used to resume a SSL session later. Example::
+
+          import socket
+          import ssl
+          addr = socket.getaddrinfo('www.google.com', 443)[0][-1]
+          sock_one = ssl.wrap_socket(socket.socket())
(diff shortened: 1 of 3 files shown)


husigeza:
Thanks for the contribution, we checked this PR and approved, it will be part of the next releases.

PR #394, opened 2020-02-04
MicroPython release 1.20.2.rc2 based on Espressif IDF 3.3.1

Xykon:


husigeza (approved):
I think the PR should be renamed, but the content is good.

PR #401, opened 2020-02-16
Fixing crash when make_new is called with zero arguments (reopen)

knagymate:
Define uart id as int instead of mp_obj.


peter-pycom:
internal PR: https://github.com/pycom/Firmware-Development/pull/226
closing this one

PR #413, opened 2020-02-26
Fix integer overflow in filesystem stat implementation

### LINKED ISSUE #409: Integer overflow in filesystem stat implementation
Dear Pycom team,

while investigating an issue why `rshell` would upload our files over and over again despite its `rsync` implementation should compensate for that (https://github.com/dhylands/rshell/issues/124), we just found an issue when looking at `mtime` timestamps of the files created on the device.

We found that there might be an integer overflow within the virtual filesystem implementation keeping track of filesystem metadata.

This example says a thousand words:
```
>>> import time; f = open('/flash/test', 'w'); f.write('hello'); f.close(); time.time() - os.stat('/flash/test')[7]
2147483648
```

It looks like only newer firmware releases are affected. The bug apparently started appearing in `1.20.0.rc0` and is still visible in `1.20.2rc3-pybytes`. However, it is not present in `1.18.2.r7` and `1.18.3-pybytes`. It looks like it is independent of FatFS vs. LittleFS.

`2147483648` is just `2^31` or `INT_MAX + 1` or `abs(INT_MIN)`. The correct outcome should be zero (`0`), or sometimes `1` - dependent on how fast the oneliner is executing.

Thanks already for looking into this.

With kind regards,
Andreas.

cc @husigeza

amotl:
Dear Pycom team,

this fixes #409, kudos go to @robert-hh. Thanks a bunch!

With kind regards,
Andreas.

### DIFF
--- a/esp32/littlefs/vfs_littlefs.c
+++ b/esp32/littlefs/vfs_littlefs.c
@@ -686,9 +686,9 @@ STATIC mp_obj_t littlefs_vfs_stat(mp_obj_t vfs_in, mp_obj_t path_param) {
     else {
         t->items[6] = mp_obj_new_int_from_uint(0); // st_size
     }
-    t->items[7] = MP_OBJ_NEW_SMALL_INT(seconds); // st_atime
-    t->items[8] = MP_OBJ_NEW_SMALL_INT(seconds); // st_mtime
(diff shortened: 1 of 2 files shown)


amotl:
Apparently, this has just been committed by @geza-pycom and merged by @Xykon through b74db42. So, I am deliberately closing this.

Issue #411, opened 2020-02-25
Can't send long AT commands using lte.send_at_cmd


leroxyl:
* GPy + pytrack or pysense
* (sysname='GPy', nodename='GPy', release='1.20.1.r2', version='v1.11-06dfad0 on 2019-11-30', machine='GPy with ESP32', pybytes='1.3.0')

Hello,

for my application I need to send long AT+CSIM commands (over 500 bytes) to the modem. Using `lte.send_at_cmd` with commands that size causes a crash. I followed the source code through the rabbit hole and found that the buffer behind this method is only 127 bytes big and a carriage return is automatically appended to the command here: https://github.com/pycom/pycom-micropython-sigfox/blob/Release/v1.20/esp32/lte/lteppp.c#L662 which hinders me from sending my command in several chunks.

Is there maybe another way to safely write to the UART without a carriage return being added?

Thank you!

robert-hh:
As far as I understand the cited code, the CR is only sent if the command is "+++".

leroxyl:
I fell for that, too. It looks counter-intuitive but `strcmp` returns 0 if the contents of both strings are equal and 0 evaluates to `false`.

robert-hh:
Oops. You are right.

razvandragomirescu:
I am experiencing the same issue, we are sending long command APDUs to the SIM card via AT+CSIM and we run out of space for some commands. This issue is almost 2 years old now, could you please fix it?

PR #414, opened 2020-02-27
esp32/lteppp: Don't send CR each time when invoking "lteppp_send_at_cmd"

amotl:
@LaRox11 found this when trying to send AT+CSIM commands longer than 500 bytes to the modem. As the buffer behind this method is only 127 bytes long and the strcmp was evaluated erroneous, this stopped users sending longer commands in several chunks.

See also #411.


robert-hh (changes requested):
This change is wrong, since the +++ sequence must not be followed by NL. After +++ no other bytes must follow for a certain   time. As far as I recall, for one second. The only appropriate change would be to allow for longer messages. But that is a larger change, affecting the buffer handling.

amotl:
Thanks, @robert-hh. I will retract this change and close the PR.

Issue #111, opened 2017-12-22
Timer.Alarm does not immediately update free timer count after alarm.callback(None) 


rkeiii:
Please include the following information when submitting a bug report:

* The Pycom board you are using: WiPy 2.0
* The firmware version you are using: custom build of 1.11.0.b1
```
Version Information:  (sysname='WiPy', nodename='WiPy', release='1.11.0.b1', version='68057a7 on 2017-12-22', machine='WiPy with ESP32')
```

* Exact steps to cause this issue
    1. Create 16 Timer.Alarm objects 
    2. Call alarm.callback(None) on several of them in rapid succession
* What you expected
    * I expected to no longer receive "MemoryError: maximum number of 16 alarms already reached" after calling alarm.callback(None) on a Timer.Alarm object 
* What happened instead
    * I received "MemoryError: maximum number of 16 alarms already reached" for some number of seconds before the Timer.Alarm code finally realized there were some free timers.

Here's the "timer code" that I'm using the instantiate the Timer.Alarm's and also to free them with alarm.callback(None).
```
import config
import util
import _thread
import time
import msg
import gc

from machine import Timer
from exception import TimersSoftLimitException, TimersHardLimitException
from uasyncio.queues import Queue
from uasyncio.queues import QueueEmpty

# TODO: This looks like insanity because it is.
# The first time you load datetime it errors with
# AttributeError: type object 'tzinfo' has no attribute '__new__'
# Upon the second load it load fine!?
try:
    from datetime import datetime
except AttributeError:
    pass
from datetime import datetime

timers = []
timers_lock = _thread.allocate_lock()
task_queue = Queue(maxsize=config.queued_tasks_limit)

class BaseTimer:
    """Base timer class"""

    def __init__(self, wakeup_callback, sleep_us, is_periodic):
        try:
            with timers_lock:
                if len(timers) < config.timers_hard_limit:
                    gc.collect()
                    self._alarm = Timer.Alarm(wakeup_callback, ms=sleep_us, arg=None, periodic=is_periodic)
                    timers.append(self)
                    print("timers in use:", len(timers))
                else:
                   raise TimersHardLimitException()
        except MemoryError as e:
            print("Probably spurious MemoryError suppressed: ", e)


class EtaTaskCallTimer(BaseTimer):
    """Timer to supprt ETA type Celery Task calls"""

    def __init__(self, task):
        # calculate how to long to sleep based on Task ETA
        now = datetime(*config.rtc.now())
[...]

PR #417, opened 2020-03-02
Forum topic 5740

Xykon:
Implemented change suggested by @robert-hh https://forum.pycom.io/topic/5740/new-firmware-release-1-20-2-rc6

### DIFF
--- a/lib/utils/pyexec.c
+++ b/lib/utils/pyexec.c
@@ -491,6 +491,7 @@ friendly_repl_reset:
         } else if (ret == CHAR_CTRL_D) {
             // exit for a soft reset
             mp_hal_stdout_tx_str("\r\n");
+#if (VARIANT == PYBYTES)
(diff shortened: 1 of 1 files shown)


catalinio (approved):
Seems ok.

DG12:
Sorry after following links above it seems that this fixes:
modpycom.c/pulses_get(): prevent lock-up at long pulses #110  and
Timer.Alarm does not immediately update free timer count after alarm.callback(None) #111

What does "**Allow** CTRL-D when using the Base VARIANT" mean?
Thank you

robert-hh:
Ctrl-D from REPL causes a soft reset. This was disabled in the cited release, even when the firmware build did not use Pybytes. This firmware variant is called BASE. This fix re-enables soft reset with Ctrl-D for the Base Variant of the firmware. This variant is not available though the Pycom Updater. You have to build it yourself from the sources.

PR #429, opened 2020-04-06
Adding support for AT commands that are longer than 124 bytes.

ftylitak:
This Pull Request adds support AT commands of arbitrary length. 

The current implementation has an AT command length limit of 124 bytes (LTE_AT_CMD_SIZE_MAX - 4) which is a blocking point when trying to use the complete AT command reference of SEQUANS modem. 

The issue has already been reported through #411 and tried to be solved by #414 though not a proper solution has been found. @robert-hh pointed out that bigger buffers would solve the problem though this would affect the buffer handling. 

My approach allows to send long AT command in chunks of max size 124. In case the command is smaller than 124 byte long, it will behave as it previously did. 

A good example of its usage would be the case of sending byte data through a UDP socket. 

```python
from network import LTE
lte = LTE()

# set the socket option to expect the TX bytes to be provided in the form of HEX values.
lte.send_at_cmd('AT+SQNSCFGEXT=1,1,0,0,0,1')

# Open UDP socket in command mode. Google DNS IPv6 address is used of demonstration
lte.send_at_cmd('AT+SQNSD=1,1,5683,"2001:4860:4860::8888",0,8888,1')

# Request the transmission of 70 bytes (which will be 140 long HEX string)
lte.send_at_cmd('AT+SQNSSENDEXT=1,70')

# 140 characters -> 70 bytes. This messages could not be sent with the default implementation.
lte.send_at_cmd('A343164303A3730313A313130303A303A303A303A32303630886368A343164303A3730313A313130303A303A303A303A32303630886368A343164303A3730313A313130303A3')

# Shutdown socket
lte.send_at_cmd('AT+SQNSH=1')
```

Also allows to pass the 'delay' parameter to the **send_at_cmd**. It is documented in https://docs.pycom.io/firmwareapi/pycom/network/lte/ though it seems the delay was never actually passed to the underlying code. 

Tested and working on gpy device.

### DIFF
--- a/esp32/lte/lteppp.c
+++ b/esp32/lte/lteppp.c
@@ -528,8 +528,9 @@ modem_init:
             xSemaphoreGive(xLTESem);
             state = lteppp_get_state();
             if (xQueueReceive(xCmdQueue, lteppp_trx_buffer, 0)) {
+                bool expect_continuation = lte_task_cmd->expect_continuation;
                 lteppp_send_at_cmd_exp(lte_task_cmd->data, lte_task_cmd->timeout, NULL, &(lte_task_rsp->data_remaining), lte_task_cmd->dataLen, lte_task_cmd->expect_continuation);
-                if(!lte_task_cmd->expect_continuation)
+                if(!expect_continuation)
                     xQueueSend(xRxQueue, (void *)lte_task_rsp, (TickType_t)portMAX_DELAY);
             }
             else if(state == E_LTE_PPP && lte_uart_break_evt)
--- a/esp32/lte/lteppp.c
+++ b/esp32/lte/lteppp.c
@@ -92,7 +92,7 @@ static bool lte_uart_break_evt = false;
  ******************************************************************************/
 static void TASK_LTE (void *pvParameters);
 static void TASK_UART_EVT (void *pvParameters);
-static bool lteppp_send_at_cmd_exp(const char *cmd, uint32_t timeout, const char *expected_rsp, void* data_rem, size_t len);
+static bool lteppp_send_at_cmd_exp(const char *cmd, uint32_t timeout, const char *expected_rsp, void* data_rem, size_t len, bool expect_continuation);
 static bool lteppp_send_at_cmd(const char *cmd, uint32_t timeout);
 static bool lteppp_check_sim_present(void);
 static void lteppp_status_cb (ppp_pcb *pcb, int err_code, void *ctx);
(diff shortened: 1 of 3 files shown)


amotl:
Thanks for solving this appropriately, @ftylitak!

cc @LaRox11

esp32/lte/lteppp.c
```diff
@@ -654,22 +658,30 @@ static bool lteppp_send_at_cmd_exp (const char *cmd, uint32_t timeout, const cha
         }
 #endif
         // flush the rx buffer first
-        uart_flush(LTE_UART_ID);
+        if(!expect_continuation || 
+            (len >= 2 && cmd[0] == 'A' && cmd[1] == 'T')) // starts with AT
+            uart_flush(LTE_UART_ID);
```
peter-pycom, line 663: {curly braces} please. there's another single line if just below that also has them, and with the two line conditional

esp32/mods/modlte.c
```diff
@@ -1091,18 +1097,46 @@ STATIC mp_obj_t lte_send_at_cmd(mp_uint_t n_args, const mp_obj_t *pos_args, mp_m
     lte_check_init();
     lte_check_inppp();
     STATIC const mp_arg_t allowed_args[] = {
-        { MP_QSTR_cmd,        MP_ARG_OBJ, {.u_obj = mp_const_none} },
+        { MP_QSTR_cmd,                   MP_ARG_REQUIRED | MP_ARG_OBJ, {.u_obj = mp_const_none} },
+        { MP_QSTR_delay,                 MP_ARG_INT,  {.u_int = 10000} }
```
peter-pycom: let's use the existing named constant that was effective before ... LTE_something_something_max

esp32/mods/modlte.c
```diff
@@ -1091,18 +1097,46 @@ STATIC mp_obj_t lte_send_at_cmd(mp_uint_t n_args, const mp_obj_t *pos_args, mp_m
...
     };
     // parse args
     mp_arg_val_t args[MP_ARRAY_SIZE(allowed_args)];
+    uint32_t argLength = MP_ARRAY_SIZE(allowed_args);
```
peter-pycom, line 1242: please set it one line earlier and use it for the creation of the array

peter-pycom (changes requested):
@ftylitak Thanks for this! 

Generally looks ok. Just some small nitpicking. And please rebase on latest Dev

ftylitak:
@peter-pycom thank you for your review. 

I have made the changes in the branch and I hope that I have addressed your comments. 

It am a bit pussled though because I have pushed the commits in my branch and those commits are not shown here. Normally those are updated automatically in the PR. Any way, I will look it up.

peter-pycom:
We just merged it internally, so I'm closing this PR. It will be released with next release candidate. Thanks again for the PR @ftylitak !

razvandragomirescu:
We are experiencing this issue with the latest released firmware, could this please be merged? What is missing? We are sending long APDUs to the SIM via AT+CSIM and it appears that commands of less than 128 bytes work fine, commands over 128 bytes are silently ignored and the response of the previous AT command is returned as the response (very strange) and commands that are exactly 128 bytes appear to freeze the modem (it never replies).

PR #432, opened 2020-04-08
telnet/telnet.c: Fix handling of special Ctrl-Characters

robert-hh:
Telnet will now like calls to sys.stdin.read() consider the
setting for interrupt characters as requested by calling
the micropython.kbd_intr() call.
Ctrl-F, if active, will now perform a reboot into safe mode,
        and will be ignored if set so.
Ctrl-D at the start of a line will cause a soft reboot, albeit
       closing the telnet session.
Ctrl_C will still be checked in commands which use mp_readline(), like
the input() statement. Test script for this PR:

```
from micropython import kbd_intr
import sys

def getline(prompt):
    l = ""
    print(prompt, end="")
    while True:
        c = sys.stdin.read(1)
        if c == '\n' or c == '\r':
            print()
            break
        else:
            sys.stdout.write(c)
            l += c
    return l

def run(msg):
    print(msg)
    while True:
        try:
            l = getline("get a line: ")
            if l == "q":
                break
        except KeyboardInterrupt:
            print("Interrupted by Ctrl-C")
            break

kbd_intr(-1)
run("Full disabled")

kbd_intr(3, -1)
run("Ctrl-F disabled")

kbd_intr(-1, 6)
run("Ctrl-C disabled")
```

### DIFF
--- a/esp32/mptask.c
+++ b/esp32/mptask.c
@@ -48,6 +48,7 @@
 #include "modled.h"
 #include "esp_log.h"
 #include "mods/pybflash.h"
+#include "telnet/telnet.h"
 
 #if defined (LOPY) || defined (LOPY4) || defined (FIPY)
 #include "modlora.h"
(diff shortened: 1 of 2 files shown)


robert-hh:
Not fixed in 1.20.3.b3

robert-hh:
Closing, since the product is not supported any more.

Issue #449, opened 2020-05-19
BLE CHAR_NOTIFY_EVENT callback with same value (but must be different)


herrfrei:
Board: WiPy 2.0
Firmware: 1.20.2.rc7

I'm trying to read data from a battery management system. This has a TX characteristic where one can write commands to it and a RX where the client gets the answer. The protocol starts with 0xDD and ends a message with 0x77. The info message I want to read is sent in two sub-packages and I get two callback calls. But both contain the second part of the message (ending with 0x77). This occurs all the time when I call the info message command.

Callback is registered with:
```python
char.callback(trigger=Bluetooth.CHAR_NOTIFY_EVENT, handler=self._rx_callback)
```

And the callback itself is:

```python
def _rx_callback(self, char=None):
    data = char.value()
    print('_rx_callback data={}'.format(hex(data)))
```

I call a function every 10 seconds that outputs the name and writes the "get info command" to the device:

```python
def print_data(self):
    if self.OD[0x1800][0x2A00]['value']:
        print('Name: %s' % self.OD[0x1800][0x2A00]['value'])
    self.bms_get_base_info()
```

And here is the output I get:

```
Name: DL1907xxxx<\r><\n>
_rx_callback data=000019130304010B68FCE277<\r><\n>
_rx_callback data=000019130304010B68FCE277<\r><\n>
```

Thank you!

herrfrei:
Here are some further investigations:

The received data are copied to a char_obj's value field and the pointer to this object is queued in modbt.c:

````c
case ESP_GATTC_NOTIFY_EVT: {
        bt_char_obj_t *char_obj;
        char_obj = find_gattc_char (p_data->notify.conn_id, p_data->notify.handle);
        if (char_obj != NULL) {
            // copy the new value into the characteristic
            memcpy(&char_obj->value, p_data->notify.value, p_data->notify.value_len);
            char_obj->value_len = p_data->notify.value_len;

            // register the event
            if (p_data->notify.is_notify) {
                char_obj->events |= MOD_BT_GATTC_NOTIFY_EVT;
            } else {
                char_obj->events |= MOD_BT_GATTC_INDICATE_EVT;
            }
            if ((char_obj->trigger & MOD_BT_GATTC_NOTIFY_EVT) || (char_obj->trigger & MOD_BT_GATTC_INDICATE_EVT)) {
                mp_irq_queue_interrupt_non_ISR(gattc_char_callback_handler, char_obj);
            }
        }
[...]

PR #453, opened 2020-06-09
Fix for speedy Bluetooth NOTIFY events (subsequent call will overwrit…

herrfrei:
Problem: subsequent call will overwrite data of first call
See [issue 449](https://github.com/pycom/pycom-micropython-sigfox/issues/449)
Changes:
* Added a define to switch on/off this patch
* gattc_events_handler(): dynamic bt_char_obj_t generation on ESP_GATTC_NOTIFY_EVT
* gattc_char_callback_handler(): release of generated bt_char_obj_t

### DIFF
--- a/esp32/mods/modbt.c
+++ b/esp32/mods/modbt.c
@@ -56,6 +56,9 @@
 #include "mbedtls/sha1.h"
 #include "nvs.h"
 
+// modification to copy received characteristic data for callbacks
+#define CB_ALLOCATE_CHAR_OBJ
+
 /******************************************************************************
  DEFINE PRIVATE CONSTANTS
  ******************************************************************************/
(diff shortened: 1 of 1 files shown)


geza-pycom:
I could not push my commit to your repository to update this PR for some reasons. Could you please check and test this: https://github.com/geza-pycom/pycom-micropython-sigfox/tree/PR453 
I updated you commit to be the same as what is being used in the GATTS.
Usage is the same as explained here: https://development.pycom.io/firmwareapi/pycom/network/bluetooth/gattscharacteristic/

herrfrei:
Géza, I tried your (much more professional) approach and it works like a charm. My application gets now all the data without problems. I will close this issue now.

Would you please submit your patch?

geza-pycom:
Yes, we will include it in an upcoming release, thanks for the contribution !

PR #463, opened 2020-07-17
remove typo

tiborrr:
change exp32 to esp32


peter-pycom:
You are changing 193 files ... I'm not quite sure what you are trying to do

PR #476, opened 2020-08-31
RegionCN470.c: Fix RegionCN470ChannelsRemove()

robert-hh:
This function was empty before, doing nothing but returning an
error code, which was ignored, as it seems. Now it removes
a channel from the Channels table as supposed.


peter-pycom:
some context https://forum.pycom.io/topic/6303/lopy4-change-uplink-channels-by-itself/19

robert-hh:
That's what initiated the PR for CN470. The change for EU433 appeared on the road while doing the CN470 cause analysis.

robert-hh:
Not fixed yet in 1.20.3.b3

robert-hh:
Closing, since the product is not supported any more.

PR #479, opened 2020-09-09
Prep v1.20.2.r1

peter-pycom:


Xykon (approved):

robert-hh:
Hello Peter, will you merge that into the Dev branch too? This branch was the red line in the various branches that popped up and disappeared in the last months.

peter-pycom:
Hi Robert. Unfortunately it's getting a bit complicated with our branches lately. The main reason being the size of the firmware. This particular change is of little use on pygate (which is basically what Dev is right now), so no we probably won't merge it to Dev in the very near future. It is absolutely the goal to unify that again, but it might take a while.

PR #491, opened 2020-10-29
display/ssd1306.py: Bug fix and improvement

robert-hh:
- Remove the calls to the I2C primitive methods. These are not
  supported by Pycom.
- Support more display formats.


robert-hh:
Changed in 1.20.3.b3 due to a move to MP 1.13

PR #495, opened 2020-11-10
BLE update

Xykon:


esp32/mods/modwlan.c
```diff
@@ -1264,11 +1285,25 @@ mp_obj_t wlan_deinit(mp_obj_t self_in) {
...
+        if(wlan_smart_config_enabled) {
+            // If the input parameter is not the object itself but a simple boolean, it is interpreted that
+            // wlan_deinit is called from machine_sleep()
+            if(mp_obj_get_type(self_in) == &mp_type_bool) {
+                called_from_sleep = (bool)mp_obj_get_int(self_in);
+                if(called_from_sleep == true) {
+                    // stop smart config with special event
+                    xEventGroupSetBits(wifi_event_group, ESPTOUCH_SLEEP_STOP_BIT);
+                    vTaskDelay(100/portTICK_PERIOD_MS);
+                }
```
peter-pycom, line 1300: just checking, is it intentional that we do nothing if self_in is bool and False? Or should we set ESPTOUCH_STOP_BIT?

peter-pycom (changes requested):
see one question inline. also please add a commit message that explains why we're doing this. what does it achieve and PR to FD:master

PR #509, opened 2020-12-25
modled.c, mperror.c: Ensure 280µs reset time for the RGB LED

robert-hh:
Otherwise, if two consecutive calls to pycom.rgbled() are called,
the second one may not get effective.


peter-pycom:
some context: https://forum.pycom.io/topic/6674/what-do-i-misunderstand-about-rgbled/19

robert-hh:
I consider this PR supportive - and because of that I made it. It should however be noted, that indeed the trailing of consecutive calls to pycom.rgbled() will be delayed, up to 280 µs. That is because pycom.rgbled() waits for the previous rgbled cycle to be finished. Instead one can always wait in the own code doing something reasonable instead of waiting.

PR #516, opened 2021-01-14
sflash_diskio_littlefs.c: Enable block level wear-leveling

robert-hh:
Otherwise, superblocks will be overly used and wear out in cases,
where a single file is rewritten again and again.

See https://forum.pycom.io/topic/6734/external-flash-lose-files

### DIFF
--- a/esp32/littlefs/sflash_diskio_littlefs.c
+++ b/esp32/littlefs/sflash_diskio_littlefs.c
@@ -45,7 +45,7 @@ struct lfs_config lfscfg =
     .prog_size = SFLASH_BLOCK_SIZE,
     .block_size = SFLASH_BLOCK_SIZE,
     .block_count = 0, // To be initialized according to the flash size of the chip
-    .block_cycles = 0, // No block-level wear-leveling
+    .block_cycles = 1000, // Enable block-level wear-leveling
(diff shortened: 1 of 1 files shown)


geza-pycom:
Hello!
May I ask why did you select the value 1000 for block_cycles? Is it based on results of your testing, or just based your best feelings?
Thanks in advance !

robert-hh:
The littlefs lib recommends values between 100 and 1000.  After short consideration it seemed to me, that for larger (free) flash sizes larger values should be taken. But I did not make any simulation. And I did not scan the internet for any resources. Obviously, any setting different to 0 is an improvement. My gut feeling would be the square root of the number of free blocks. That would be about 45. And smaller numbers have a worse performance that larger numbers, because of the additional super block swapping once that setting is reached. Micropython.org uses 100.
This setting affects the wear leveling used for the super blocks. Ordinary data block numbers are used in a round robin fashion anyhow.

geza-pycom:
Thanks for the detailed asnwer and all the analysis you have done in this topic. I would recommend to go with value 100, same as the original MicroPython project.

robert-hh:
Fixed in 1.20.3.b3

PR #520, opened 2021-01-21
LTE: fix conditional in lte_check_attached

sven-hm:
As `CEREG` was set to `1` in this [commit](https://github.com/pycom/pycom-micropython-sigfox/pull/499/commits/8e631efefb6b3d7be6731f826c06dcaf85231785) the answer to an `"AT+CEREG?"` will only contain information up to the  `<stat>`-value. So `lte_check_attached` with `legacy=true` and when `"AT+CGATT?"` returns `0` will always return false.

This tiny PR should fix the problem.

### DIFF
--- a/esp32/mods/modlte.c
+++ b/esp32/mods/modlte.c
@@ -280,8 +280,7 @@ static bool lte_check_attached(bool legacy) {
                     mp_hal_delay_ms(LTE_RX_TIMEOUT_MIN_MS);
                     lte_push_at_command("AT+CEREG?", LTE_RX_TIMEOUT_MIN_MS);
                 }
-                if (((pos = strstr(modlte_rsp.data, "+CEREG: 1,1")) || (pos = strstr(modlte_rsp.data, "+CEREG: 1,5")))
-                        && (strlen(pos) >= 31) && (pos[30] == '7' || pos[30] == '9')) {
+                if ((pos = strstr(modlte_rsp.data, "+CEREG: 1,1")) || (pos = strstr(modlte_rsp.data, "+CEREG: 1,5"))) {
                     attached = true;
                 }
             } else {


peter-pycom:
You're completely right. Yes, cereg 1 broke this line and I think it's safe to remove these checks

Thanks. We'll include it in the next release.

I see you set it to Draft, were you thinking of adding other changes?  ... 

Also, I would like to better understand the background - did you actually get a lte.isattached() False even though you were attached according to CEREG? If so, you must have gotten a CGATT 0. And I'd be very interested in the details, because I have never seen that. Are you using any additional AT commands? carrier param? Which country, provider, band, FW version, etc are you using?

sven-hm:
Actually i first wanted to understand the difference between getting the attached-state information from the `AT+CGATT?` and `AT+CEREG?`. But if you say the fix doesn't break anything, i'll un-draft it :)
I had some debug-output in those lines and saw the case that `AT+CGATT?` returned 0 and `AT+CEREG?` returned `+CEREG: 1,5`, we have the `lte.isattached()` in a retry loop, so it took some retries and `AT+CGATT?` returned 1. I tried to reproduce this today in the morning but couldn't.
_Before_ we try to connect we only use `AT+CIMI` to read the IMSI.

Some more details:
* country: Germany
* provider apn: iot.1nce.net
*  band: 8
* board: GPy 1.0
*  FW version: v1.20.2.r4
*  modem firmware: UE5.0.0.0d, LR5.1.1.0-41065

peter-pycom:
Mhm. Could it be that it was a race condition? I mean if it only happened once, it could be that at the time of cgatt it was not attached, but a few milliseconds later at the time of cereg, it was attached, but it tested for a cereg=2 return value. That failed so it returned False. The subsequent call to isattached() should then return True, because now both cgatt and cereg both show that its attached.

Could it be like that?

sven-hm:
That might be possible. I left one regularly attaching device running over the last days with some debug logging and didn't have the case again.

Issue #523, opened 2021-01-31
LFS: Closing a file twice crashes the firmware


robert-hh:
Version 1.20.2.r4, although I guess that this does not matter
The three code lines below cause a crash by abort() being called by RTOS.
The same code snippet behaves well with FAT, behaves well with LFS on micropython.org firmware. 
A dir(f) call after the first f.close() causes the firmware to freeze. On mp.org firmware, it still shows the file methods. That indicates that the file object gets corrupted on close. Until now, I did not find the place where that happens.
Even if closing a file twice is not reasonable, the firmware should be able to deal with it, either by ignoring or raising an error.
Code lines:
```
f = open("main.py")
f.close()
f.close()
```
start of decoded trace dump:
```
BT-0: 0x40090638 is in invoke_abort (/home/robert/Downloads/MicroPython/pycom-esp-idf/components/esp32/panic.c:156).
156	        *((int *) 0) = 0;
BT-1: 0x40090885 is in abort (/home/robert/Downloads/MicroPython/pycom-esp-idf/components/esp32/panic.c:171).
171	    invoke_abort();
BT-2: 0x40098170 is in multi_heap_free_impl (/home/peter/docs/pycom-esp-idf/components/heap/multi_heap_platform.h:54).
54	        abort();
BT-3: 0x400852b2 is in heap_caps_free (/home/peter/docs/pycom-esp-idf/components/heap/heap_caps.c:268).
268	    multi_heap_free(heap->heap, ptr);
BT-4: 0x40084dc5 is in _free_r (/home/peter/docs/pycom-esp-idf/components/newlib/syscalls.c:42).
42	    heap_caps_free( ptr );
BT-6: 0x400f7e2a is in lfs_file_rawclose (littlefs/lfs_util.h:226).
226	    free(p);
BT-7: 0x400f89c5 is in lfs_file_close (littlefs/lfs.c:5089).
5089	    err = lfs_file_rawclose(lfs, file);
BT-8: 0x400f9676 is in littlefs_close_common_helper (littlefs/vfs_littlefs.c:326).
326	    int res = lfs_file_close(lfs, fp);
BT-9: 0x400f973f is in file_obj_ioctl (littlefs/vfs_littlefs_file.c:105).
105	            int res = littlefs_close_common_helper(&self->littlefs->lfs, &self->fp, &self->cfg, &self->timestamp_update);
BT-10: 0x4010c97d is in mp_stream_close (../py/stream.c:422).
[...]

PR #524, opened 2021-02-01
vfs_littlefs_file.c: Prevent double close of a file

robert-hh:
An attempt to close a file twice resulted in a crash, because buffers
should have been freed again. This change checks first, if a file is
already closed.

### DIFF
--- a/esp32/littlefs/vfs_littlefs_file.c
+++ b/esp32/littlefs/vfs_littlefs_file.c
@@ -100,17 +100,19 @@ STATIC mp_uint_t file_obj_ioctl(mp_obj_t o_in, mp_uint_t request, uintptr_t arg,
         return 0;
 
     } else if (request == MP_STREAM_CLOSE) {
+        if (self->littlefs == NULL) {
+            return 0;
+        }
 
         xSemaphoreTake(self->littlefs->mutex, portMAX_DELAY);
             int res = littlefs_close_common_helper(&self->littlefs->lfs, &self->fp, &self->cfg, &self->timestamp_update);
         xSemaphoreGive(self->littlefs->mutex);
+
+        self->littlefs = NULL; // indicate a closed file
         if (res < 0) {
             *errcode = littleFsErrorToErrno(res);
             return MP_STREAM_ERROR;
         }
-        // Free up the object so GC does not need to do that
-        m_del_obj(pyb_file_obj_t, self);
-
         return 0;
     } else {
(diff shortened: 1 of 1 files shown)


peter-pycom:
https://github.com/pycom/pycom-micropython-sigfox/issues/523

geza-pycom:
Hello!
Good catch, it was hard to decode because without removing the  _m_del_obj(pyb_file_obj_t, self)_ call the exception is dropped by _mp_load_method_maybe_().
I checked your example the issue happens and your fix solves it.

However with the following example, without your fix, the problem does not come alive:
`
>>> f = open("myfile.txt", "w")
>>> f.write("ssss")
4
>>> f.close()
>>> f.close()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'function' object has no attribute 'close'
>>> `

I think here the exception is correct, because the file object ("f") is deleted by _m_del_obj(pyb_file_obj_t, self)_.

It would be interesting to see why the crash happens with your example and why it does not with this other example.

esp32/littlefs/vfs_littlefs_file.c
```diff
@@ -100,17 +100,19 @@ STATIC mp_uint_t file_obj_ioctl(mp_obj_t o_in, mp_uint_t request, uintptr_t arg,
...
     } else if (request == MP_STREAM_CLOSE) {
+        if (self->littlefs == NULL) {
+            return 0;
+        }
 
         xSemaphoreTake(self->littlefs->mutex, portMAX_DELAY);
             int res = littlefs_close_common_helper(&self->littlefs->lfs, &self->fp, &self->cfg, &self->timestamp_update);
         xSemaphoreGive(self->littlefs->mutex);
+
+        self->littlefs = NULL; // indicate a closed file
```
geza-pycom: I think this should be moved after checking the return value, because if the return value is smaller than 0 then an error occured which means the file has not been closed. I would move this to line 111.

robert-hh:
I'm a little bit reluctant to delete python objects deep inside drivers, unless you know that they are not used any more. When I wrote that, I had a clear opinion why it was wrong here, but now I do not know it any more. Anyhow, it does not to keep it and let the garbage collector do the job later.

geza-pycom:
Thanks, this fix looks good !

robert-hh:
Fixed in 1.20.3.b3

Issue #518, opened 2021-01-20
LFS: Core dumps on heavy file operations


robert-hh:
Firmware: 1.20.2.r4, WiPy3, with a change for block level wear leveling and RGB led guard.
The error happens during a long term test, aiming at writing the same short file several million times, in order to verify the wear leveling change.
At random intervals, about every 50_000 file open/write/close cycles, a core dump happens. It may happen after a few thousand cycles, or after 100_000 cycles.
Test code:
```
import time
line = "*" * 63 + "\n"
b = 31 * line
        
def write(name, n):
    f = open(name, "w")
    f.write("%8d\n" % n)
    f.write(b)
    f.close()

def run1(start=1, end=10_000_000):
    try:
        f = open("logfile")
        l = f.readline()
        f.close()
        n = int(l.strip())
    except:
        n = start
    print(n, "Restart ")

    while n < end:
        print(n, "--- ")
        write("logfile", n)
        n += 1
        time.sleep(0.1)
```
The backtrace of a typical core dump looks like. I decoded 6 core dumps. They all happen on a file close. And the exception always happens at the rmt_tx interrupt, which is served by Core1. RMT is activated by the heartbeat in the main loop, which is served by Core 0. The backtrace of Core shows, that at the time of the heartbeat the core 0code is working on sending a RGB led flash. Whether still or again, is unclear.
```
Guru Meditation Error: Core  1 panic'ed (Interrupt wdt timeout on CPU1)
Core 1 register dump:
PC      : 0x4009602e  PS      : 0x00060034  A0      : 0x800952ec  A1      : 0x3ffbf4f0  
A2      : 0x3ffe2b4c  A3      : 0x00060023  A4      : 0x00000001  A5      : 0x0000cdcd  
A6      : 0xb33fffff  A7      : 0x3ffdd385  A8      : 0x0000cdcd  A9      : 0x00000000  
A10     : 0x0000039f  A11     : 0x0000abab  A12     : 0x3f40df58  A13     : 0x00000000  
A14     : 0x3f40df40  A15     : 0x01f12318  SAR     : 0x00000003  EXCCAUSE: 0x00000006  
EXCVADDR: 0x00000000  LBEG    : 0x40093a80  LEND    : 0x40093a93  LCOUNT  : 0x00000000  
Core 1 was running in ISR context:
EPC1    : 0x400f5227  EPC2    : 0x00000000  EPC3    : 0x00000000  EPC4    : 0x4009602e

[...]

PR #525, opened 2021-02-01
mperror.c: Move noticing of the next heartbeat transition

robert-hh:
The place at which the time for the next transition is noticed
moves from the start of the respective block to it's end, when
the RGB led has switched.
The effect on heartbeat timing is minor. Without load, the heartbeat 'on' and 'off' 
times are identical for the old and new version. Timing with load new version. The first pulse
at -200ms is the 'on' command for the RGB, the scattered second pulses are the 'off' commands accumulated over 12 hours.

heartbeat_timing_new_w_load

Timing with load old version.
heartbeat_timing_orig_w_load

### DIFF
--- a/esp32/util/mperror.c
+++ b/esp32/util/mperror.c
@@ -131,16 +131,18 @@ bool mperror_heartbeat_signal (void) {
         mperror_heart_beat.do_disable = false;
     } else if (mperror_heart_beat.enabled) {
         if (!mperror_heart_beat.beating) {
-            if ((mperror_heart_beat.on_time = mp_hal_ticks_ms_non_blocking()) - mperror_heart_beat.off_time > MPERROR_HEARTBEAT_OFF_MS) {
+            if (mp_hal_ticks_ms_non_blocking() > mperror_heart_beat.off_time) {
                 led_info.color.value = MPERROR_HEARTBEAT_COLOR;
-                led_set_color(&led_info, false, false);
+                led_set_color(&led_info, true, false);
                 mperror_heart_beat.beating = true;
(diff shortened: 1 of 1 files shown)


peter-pycom:
some context: 

https://forum.pycom.io/topic/6734/external-flash-lose-files/87?_=1615194014360

> b) During that wear levelling test, I had core dumps at about every 50_000 writes. They were related to the heartbeat flash. I made a PR which changed that heartbeat timing a little bit. After that, the crashes disappeared, at least in that test. I am not 100% confident that the change cured the initial problem, which could also be a Core-0/Core-1 collision. But the change is also not intrusive and saves a few clock cycles. PR here: https://github.com/pycom/pycom-micropython-sigfox/pull/525

geza-pycom:
Hello!
Could you please detail why this fix is needed? What issue does this fix? Thanks !
Is it related to https://github.com/pycom/pycom-micropython-sigfox/issues/518 ?

robert-hh:
> Is it related to #518 ?

Yes. Analyzing the backtraces of this fails, it looked like a race condition which involved the code doing the heartbeat and the code, closing a file. The previous code reported a new time before the actual color change was done. The change moves registering the new time after the color has changed. Before the change, the code would crash about every few hours in average, After that, the code ran fine for a week, after which is was stopped manually. It was the wear leveling test. The device with this change code did 4 Million cycles. another one 10 Million file create/write/closes. 
And, as a slight performance improvement,  it performs the assignment and subtraction only when needed. I am definitely not sure about the mechanism. So the change may only hide the real reason.

P.S.: I had another crash case caused by the heartbeat led, which disappeared after I disabled it. But I did not look into it yet.

geza-pycom:
Thanks for the contribution, this change will be part of an upcoming release.

PR #544, opened 2021-04-28
Adding support for AT commands that are longer than 124 bytes

ftylitak:
Based on the conversation of #429 , the pull request was merged. 

Though after testing, a specific code part is missing thus making the feature unusable. 

In case you are interested in having this feature, this commit adds the missing part for the support of long AT command requests.


razvandragomirescu:
We are experiencing this issue with the latest released firmware, could this please be merged? What is missing? We are sending long APDUs to the SIM via AT+CSIM and it appears that commands of less than 128 bytes work fine, commands over 128 bytes are silently ignored and the response of the _previous_ AT command is returned as the response (very strange) and commands that are _exactly_ 128 bytes appear to freeze the modem (it never replies).

